PR #26, opened 2020-12-27
Quality of Life improvements

JElgar:
There a lot of stuff on my branch atm which should really be in master. Ill merge this in in a bit with some of those things


JElgar:
Change of plan just gonna merge in create bookings early

PR #38, opened 2021-01-06
Updated Makefile with user for migrations

JElgar:
LGTM


streth11 (approved):
Now we're cooking with gas

PR #62, opened 2021-01-26
Feature/address gql

streth11:


uobtheatre/venues/test/test_schema.py
```diff
@@ -22,6 +22,16 @@ def test_venues_schema(gql_client, gql_id):
...
                 name
+                address {
+                  id
+                  buildingName
+                  buildingNumber
+                  street
+                  city
+                  postcode
+                  latitude
+                  longitude
```
JElgar, line 33: Sorry @streth11 I have misled you. Keep just the id in here to make sure that the address is accessible. Then in addresses/test/test_schema add all the fields in an AddressNode. You will need to access the address through a venue in that new test but that is fine.

JElgar (changes requested):

uobtheatre/venues/test/test_schema.py
streth11: i think i have fixed

uobtheatre/addresses/test/test_schema.py
```diff
@@ -0,0 +1,58 @@
+import pytest
+
+# from uobtheatre.addresses.test.factories import PerformanceFactory
+from uobtheatre.venues.test.factories import VenueFactory
+
+
+@pytest.mark.django_db
+def test_venues_schema(gql_client, gql_id):
```
JElgar, line 8:
```suggestion
def test_address_schema(gql_client, gql_id):
```

uobtheatre/venues/schema.py
```diff
@@ -19,6 +20,7 @@ class Meta:
 
 class VenueNode(GrapheneImageMixin, DjangoObjectType):
     image = GrapheneImageField(GrapheneImageFieldNode)
+    address = graphene.Field(AddressNode)
```
JElgar, line 23: In theory, this should happen automagically
```suggestion
```

JElgar (commented):
Two more small things, you can probably just accept the recommendation in GitHub and check the tests still pass

uobtheatre/venues/schema.py
JElgar: For context, it seems that `DjangoObjectType` will automatically add any field which is either a simple type (like a string or int) or if its a custom model it will add it if a `DjangoObjectType` exists for that model. So in theory simply creating the `AddressNode` will make the address available

uobtheatre/addresses/test/test_schema.py
streth11: yes

JElgar (approved):

JElgar (approved):
LGTM

PR #66, opened 2021-01-31
Added slug production and venue to schema

JElgar:


JElgar (commented):
Need to also do with id but find for now

PR #67, opened 2021-02-01
Feature/add society slug

streth11:


JElgar (changes requested):
Looks great can you also chuck in an individual resolver to the schema. 

Should be similar to this example.
https://github.com/BristolSTA/uobtheatre-api/blob/3241ea52f765432b2e1356bd39df48615f8f7aee/uobtheatre/venues/schema.py#L42

JElgar (approved):
LGTM

PR #71, opened 2021-02-03
Feature/filtering

JElgar:
- Added filtering to productions and performances
- Added tickets to performances

### DIFF
--- a/uobtheatre/productions/schema.py
+++ b/uobtheatre/productions/schema.py
@@ -1,7 +1,12 @@
+import django_filters
 import graphene
 from graphene import relay
 from graphene_django import DjangoListField, DjangoObjectType
-from graphene_django.filter import DjangoFilterConnectionField
+from graphene_django.filter import (
+    DjangoFilterConnectionField,
+    GlobalIDFilter,
+    GlobalIDMultipleChoiceFilter,
+)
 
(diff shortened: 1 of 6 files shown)


uobtheatre/productions/schema.py
```diff
@@ -1,7 +1,12 @@
+import django_filters
 import graphene
 from graphene import relay
 from graphene_django import DjangoListField, DjangoObjectType
-from graphene_django.filter import DjangoFilterConnectionField
+from graphene_django.filter import (
+    DjangoFilterConnectionField,
+    GlobalIDFilter,
+    GlobalIDMultipleChoiceFilter,
```
JElgar, line 8: We arent using this atm but I think it could be quite handy in some case?

uobtheatre/productions/schema.py
```diff
@@ -53,6 +58,18 @@ class Meta:
         interfaces = (relay.Node,)
 
 
+class ProductionFilter(django_filters.FilterSet):
+
+    start = django_filters.DateTimeFilter(method="start_filter")
+    id = GlobalIDFilter()
```
JElgar, line 64: The secret sauce. Might be worth creating a custom FilterSet class that adds this automatically because we will always want to be able to filter on id

PR #78, opened 2021-02-20
REST in peace

JElgar:
- Remove all django rest framework
- Add graphql auth
- Added error handling for mutations
- Removed pipenv


streth11 (approved):
Love It

schema.graphql
```diff
@@ -108,17 +134,68 @@ type MiscCostNode implements Node {
...
+  passwordSet(token: String!, newPassword1: String!, newPassword2: String!): PasswordSet
+  passwordChange(oldPassword: String!, newPassword1: String!, newPassword2: String!): PasswordChange
+  updateAccount(firstName: String, lastName: String): UpdateAccount
+  archiveAccount(password: String!): ArchiveAccount
+  deleteAccount(password: String!): DeleteAccount
+  sendSecondaryEmailActivation(email: String!, password: String!): SendSecondaryEmailActivation
+  verifySecondaryEmail(token: String!): VerifySecondaryEmail
+  swapEmails(password: String!): SwapEmails
+  removeSecondaryEmail(password: String!): RemoveSecondaryEmail
+  tokenAuth(password: String!, email: String): ObtainJSONWebToken
```
atoff, line 152: Thought this was changing to `login`?

atoff: Also email should definitely be required

schema.graphql
```diff
@@ -108,17 +134,68 @@ type MiscCostNode implements Node {
   value: Float
 }
 
+type Mutation {
+  register(email: String!, password1: String!, password2: String!): Register
```
atoff, line 138: Want to also collect first and last name here?

atoff: Also... If we are verifying emails, we shouldn't be able to get a token from the register mutation

uobtheatre/utils/exceptions.py
```diff
@@ -1,42 +1,78 @@
...
+# TODO We should probably raise these error instead and then format for you in
+# the mutation, ie in mutation mixin wrap the mutation funciton in a try and
+# handle errors there
+# class GQLFieldError(Exception):
+#     """
+#     A single GQL Field error
+#     """
+#
+#     def __init__(self, message, field=None, code=None, params=None):
+#         super().__init__()
```
atoff, line 13: Assuming all these comments should go too?

atoff (changes requested):
Need to have a think about how we implement all these auth functions.
Particularly:
- Have we tried out how it sends emails in response to events like registration and password resets
- Do we need to configure the emails so that the links go the right place (i.e. the password reset on the front end it /login/forgot?resetToken=<resetTokenHere>). Need to do the same with the registration verification email

schema.graphql
```diff
@@ -297,18 +413,36 @@ type TicketNodeEdge {
...
-  username: String!
   firstName: String!
   lastName: String!
-  email: String!
   isStaff: Boolean!
   isActive: Boolean!
   dateJoined: DateTime!
   id: ID!
+  email: String!
   bookings(offset: Int, before: String, after: String, first: Int, last: Int, bookingReference: UUID, user: ID, performance: ID, status: String, id: ID): BookingNodeConnection!
```
atoff, line 431: status here should be of the BookingStatus enum type

uobtheatre/utils/exceptions.py
JElgar: Yeah probably, but there are going to come back I think. The plan is to have all things in a try-catch and then throw these kinds of errors where possible

JElgar:
The whole email side of things is going to be in different ticket(s) cause I don't really think we need that yet. Im guessing you wont need that for the mocks?

schema.graphql
atoff: For all other filtering I have been doing, it is using the enum type, not a string. I assumed this would be how it works because one a booking node, the status is of the enum type, and not a string.. bit of a double standard?

i.e. 
```
bookings(performance: $performanceID, status: INPROGRESS) {
      edges {
        node {
          ...DetailedBookingDetails
        }
      }
    }
```

If this isn't possible and we have to use strings instead I guess its fine, just a bit whack switching between the gql enum types and strings

Re your point number 2, I think this is probably expected because this is not a validation error (i.e. its more a syntax error with the query we submitted to graphql) so I think thats fine.

config/settings/common.py
```diff
@@ -208,9 +208,8 @@
         "status__verified": ["exact"],
         "status__secondary_email": ["exact"],
     },
-    "REGISTER_MUTATION_FIELDS": {
-        "email": "String",
-    },
+    "REGISTER_MUTATION_FIELDS": ["email"],
+    "REGISTER_MUTATION_FIELDS_OPTIONAL": ["first_name", "last_name"],
```
atoff, line 212: These should be mandatory?

atoff:
> The whole email side of things is going to be in different ticket(s) cause I don't really think we need that yet. Im guessing you wont need that for the mocks?

This is fine. I reckon the main outstanding points are if we can rename tokenAuth to login (or if we should leave as is I guess) and fixing some of the fields to be mandatory. Can you also check what the behaviour is like for attempting to login with an account that hasn't verified their email?

schema.graphql
JElgar: Ah I see what you are saying, but when you actually get the response of the enum it should be like this: 
```
{
  "node": {
    "status": "INPROGRESS"
  }
}
```
i.e. it will also be a string

JElgar:
The login behaviour stuff should definitely be a problem for after we sorted out the emails. Other than that how do you now feel about this one @atoff

JElgar:
Added issue #78 for the email and account verification stuff.

atoff (approved):

PR #84, opened 2021-03-09
Feature/better booking refs

### LINKED ISSUE #37: Bookings update
- Change the booking reference to not be horrible (maybe some 10 digit varchar thing)
- Write function to generate booking reference
- Generate booking reference in create model
- Be able to create a qr code from the booking reference
- Add filtering to the endpoint st we can do bookings/?booking_reference=ABCDEFG or maybe even just bookings/:ref and remove /id?

streth11:
closes #37


uobtheatre/bookings/models.py
```diff
@@ -315,3 +307,23 @@ class Ticket(models.Model):
...
+        to only one ticket) applied.
+        """
+        return self.booking.performance.price_with_concession(
+            self.concession_type,
+            self.booking.performance.performance_seat_groups.get(
+                seat_group=self.seat_group
+            ).price,
+        )
+
+    def price(self):
```
JElgar, line 323:
```suggestion
    def seat_price(self):
```

uobtheatre/bookings/test/test_models.py
```diff
@@ -241,6 +240,69 @@ def test_get_price():
...
+
+
+# @pytest.mark.django_db
+# @pytest.mark.parametrize(
+#     "value, percentage, error",
+#     [(None, None, True), (None, 1, False), (1, None, False), (1, 1, True)],
+# )
+# def test_misc_cost_constraints(value, percentage, error):
+
+
```
JElgar, line 266:
```suggestion
```

requirements.txt
```diff
@@ -29,3 +29,5 @@ graphene==2.1.8
 graphene-django==2.15.0
 
 pytz==2021.1
+
+shortuuid>=1.0.1
```
JElgar, line 33: Can we add a short comment after this to say why its imported (just helps avoid clutter when we check if we can get rid of any)

JElgar (changes requested):

JElgar (approved):

PR #95, opened 2021-03-16
Added path, module and test to makefile

JElgar:


JElgar:
Requires README update

JElgar:
In reality need to rewrite the whole readme

PR #99, opened 2021-03-17
my code is basically perfect

walkerkit:
Did a quick review of the mutations, I'm happy


walkerkit:
The __eq__ and subsequent __hash__ functions haven proven too jank... the situation has become untenable! 
Society has nearly broken down, if not for this PR.

For more information see this vaguely relevant SO question: [here!](https://stackoverflow.com/questions/61212514/django-model-objects-became-not-hashable-after-upgrading-to-django-2-2)

uobtheatre/utils/test_utils.py
```diff
@@ -0,0 +1,24 @@
...
+        ticketKey = (
+            ticket.get("seat_group_id"),
+            ticket.get("concession_type_id"),
+            ticket.get("seat_id"),
+        )
+        if ticketKey in ticketDict:
+            ticketDict[ticketKey] += 1
+        else:
+            ticketDict[ticketKey] = 1
+    return ticketDict
```
JElgar, line 13: Could we do something like this?

```suggestion
    ticketObjectList = map(Ticket, ticketDictList)
    return ticketListDictGen(ticketObjectList)
```

JElgar (changes requested):
Very nice thank you Kit. Only one thing is it looks like those functions duplicate logic a bit. What do you think about something like the suggestion?

uobtheatre/utils/test_utils.py
JElgar: You might actually need to use list comp here eg:
`ticketObjectList = [Ticket(**ticketDict) for ticketDict in ticketDictList]`

walkerkit: Lovely suggestion there James!

JElgar (approved):

PR #100, opened 2021-03-18
Fix/graphql inconsistencies

### LINKED ISSUE #46: tickets_inc_discount_price
Alex says we should add `tickets_discounted_price` to the response and then `sub_total` should be reserved for the future when we have extra stuff (like programs and :popcorn: ).

DoD: 
Add extra field to ticket price break down serializer which also gives the subtotal called `tickets_discounted_price`

### LINKED ISSUE #63: Add custom image model
I think this is a good idea but not 100% yet.
Somehow extend existing image model. Then just create a schema for that new model and remove all the nasty mixin stuff.
Then we can add extra stuff onto the image model as required.

### LINKED ISSUE #98: image full URL

streth11:
fixes many of the inconsistencies with the front-end

closes #98 
closes #63 
closes #46 

TODO:

- [x] Test bookings orderBy filter
- [x] Test IdInputField with parameters
- [x] Image Model


streth11:
wait for kit's stuff before merge

walkerkit:
Are we still brining this in before the refactor?

uobtheatre/bookings/schema.py
```diff
@@ -35,25 +45,51 @@ class Meta:
...
+
+class DiscountNodeMixin:
+    discounts = DjangoListField(DiscountNode)
+
+    def resolve_discounts(self, info):
+        return [
+            discount
+            for discount in self.discounts.all()
+            if not discount.is_single_discount()
+        ]
```
JElgar, line 70: There are a few options for this which I think might be a bit betterer. One option is to set the discount node queryset but I think the ideal is to somehow set the "default" filtering.

JElgar (changes requested):

JElgar:
@walkerkit lets branch off this and then start the refactor since there will be very few changes for this branch before its merged.

uobtheatre/bookings/schema.py
streth11: james this fix is beyone my level, sooo its all yours bby

JElgar (approved):
LGTM

PR #141, opened 2021-05-20
Feature/new docker

JElgar:


JElgar:
Bit whack but more fixing can happen later

PR #178, opened 2021-07-14
Export db url in local config

JElgar:


atoff:
Does this work when to `exec bash` into to container?

JElgar:
@atoff that is what this is meant to be fixing, so hopefully...

I find it so strange that we are using a url for something and separate variables in others so we should probably just fix that now instead.

JElgar:
Temp fix pushed in #166, should switch to just using one type of environment variables.

PR #183, opened 2021-07-18
Box office bookins

JElgar:
- Allow Boxoffice admins to specify which user a booking is being created for, if user does not exist the user is created
- Add admin_discount to allow for concession booking (support in mutation still required)
- Move discounts to a separate app (tests still need moving)


schema.graphql
```diff
@@ -27,6 +27,7 @@ type BookingNode implements Node {
   user: UserNode
   performance: PerformanceNode!
   status: EnumNode
+  adminDiscount: Float!
```
atoff, line 30: If this is always percentage could be `adminDiscountPercentage`?

schema.graphql
```diff
@@ -162,7 +163,7 @@ type MiscCostNode implements Node {
 }
 
 type Mutation {
-  createBooking(performanceId: IdInputField, tickets: [CreateTicketInput]): CreateBooking
+  createBooking(performanceId: IdInputField, targetUser: String, tickets: [CreateTicketInput]): CreateBooking
```
atoff, line 166: `targetUserEmail`

atoff (changes requested):

atoff (approved):
lgtm

PR #189, opened 2021-07-20
Feature/target user for booking update

### LINKED ISSUE #187: Booking target user on update
At the moment we can only set the `target user email` on creation of the booking. This should also be available on the update mutation (the front end flow is create the booking, enter the user email, then pay)

JElgar:
closes #187

### DIFF
--- a/db.json
+++ b/db.json
@@ -457,7 +457,7 @@
     }
   },
   {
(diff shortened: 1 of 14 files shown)


atoff (approved):
🥳

PR #210, opened 2021-08-02
Skip group discount calculations when not required

JElgar:
Speed up calculation of ticket price breakdown by skipping group discount calculation when performance has no group discounts

### DIFF
--- a/uobtheatre/bookings/models.py
+++ b/uobtheatre/bookings/models.py
@@ -274,10 +274,11 @@ class Booking(TimeStampedMixin, models.Model):
         Returns:
             int: price of the booking with discounts applied in penies
         """
-        return math.ceil(
-            self.get_best_discount_combination_with_price()[1]
-            * (1 - self.admin_discount_percentage)
-        )
+        if self.performance.has_group_discounts:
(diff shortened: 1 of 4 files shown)


streth11 (commented):
looks spicy but want to check with front end before merge to see if it actually works

atoff (approved):
Looks like it should do the job

uobtheatre/bookings/models.py
```diff
@@ -274,10 +274,11 @@ def subtotal(self) -> int:
         Returns:
             int: price of the booking with discounts applied in penies
         """
-        return math.ceil(
-            self.get_best_discount_combination_with_price()[1]
-            * (1 - self.admin_discount_percentage)
-        )
+        if self.performance.has_group_discounts:
```
atoff, line 277: Actually, I feel like this logic should be on the `tickets_discounted_price` function

atoff (changes requested):

uobtheatre/bookings/models.py
JElgar: There is no such thing. Not until we support other payable objects (other than just bookings).

PR #223, opened 2021-08-05
Added tiptap

JElgar:


JElgar:
@atoff any other fields for admin pannel?

atoff (changes requested):
I think society and venue descriptions should also be tiptap-able

atoff (approved):

PR #226, opened 2021-08-09
feat: Production Status

### LINKED ISSUE #201: "Draft" performances/productions
At the moment, productions which don't have all the required defined data (particularly that don't have images / performances / seat groups/ concession types, etc) are able to be shown to generic users via queries.

We should have a system so that only authorised people are able to get productions/performances that are in draft via queries

atoff:
Implements a status field on the production model

Closes #201


atoff:
@JElgar  Can we implement the Enum node here?

uobtheatre/productions/test/test_schema.py
```diff
@@ -816,12 +818,37 @@ def test_production_filters(filter_name, value_days, expected_outputs, gql_clien
 
 
 @pytest.mark.django_db
-def test_perfromance_run_on(gql_client):
+def test_draft_productions_not_shown(gql_client_flexible):
```
JElgar, line 821:
```suggestion
def test_draft_productions_not_shown_to_basic_user(gql_client_flexible):
```

JElgar (changes requested):
Looks good but the testing seems a bit weedy. When we were on the call there were 3 cases you tested:
- Logged in as a user with some draft permissions
- Logged in as not admin
- Not logged in
I would say all these cases should be tested for us to be confident that `Production` `qs` is doing what we want.

I also notice you have changed a lot of `gql_client` to `gql_client_flexible`. I assume this is because tests started to fail with the new filtering and no user. I think a better solution that signing in the user for every production test is to set the default status of a `Production` in the `ProductionFactory` to `PUBLISHED` and then keep the user logged out

conftest.py
```diff
@@ -6,21 +6,23 @@
 from graphene.test import Client as GQLClient
 from pytest_factoryboy import register
 
-from uobtheatre.schema import schema
+from uobtheatre.schema import schema as appSchema
```
JElgar:
```suggestion
from uobtheatre.schema import schema as app_schema
```

JElgar (approved):
LGTM. Thanks for sorting the gql_client nonsense

PR #230, opened 2021-08-13
Payments refactor and square terminal support

JElgar:
- Refactored payments to use PaymentMethod abstract class
- Added `complete` method to booking
- Added square terminal payment method
- Added square terminal checkout webhook handler. This currently only supports terminal checkout events where the status is `COMPLETED`
- Removed `gql_id` (using `to_global_id` instead)
- Updated `mock_square` to use contextmanager in test


conftest.py
```diff
@@ -1,24 +1,30 @@
...
 
+@pytest.fixture
+def rest_client():
+    return APIClient()
+
+
 class AuthenticateableGQLClient(GQLClient):
-    """A graphql client extension that supports a request and user"""
+    """
+    Graphql client which can be loged in and out.
```
atoff, line 26: `logged`

conftest.py
```diff
@@ -49,34 +55,47 @@ def execute(self, query, variable_values=None):
...
-        "utf-8"
-    )
-
-
-@pytest.fixture
-def mock_square(monkeypatch):
-    """Mocks the square library"""
+def mock_square():
+    """
+    Used to mock the square clinet
```
atoff, line 60: `client`

uobtheatre/payments/payment_methods.py
```diff
@@ -0,0 +1,316 @@
...
+                info. When connection to a device the `device_id` should be
+                used.
+
+        Raises:
+            SquareException: If the square request returns an error
+        """
+        response = cls.client.devices.list_device_codes()
+        if not response.is_success():
+            raise SquareException(response)
+        return response.body["device_codes"]
```
atoff, line 200: Does this only list device codes that have been "paired"? Also must have the product code of `TERMINAL_API` or something

JElgar: I've added some filters for this, see this test for an example
https://github.com/BristolSTA/uobtheatre-api/blob/0528b39f108f07dc80f48ac5336b7bb45a71475f/uobtheatre/payments/test/test_schema.py#L214

atoff: Yes, but I'm wondering if we actually want to expose all these? Other device codes are only used for the POS app (I think). I don't think the front end should be concerned about having to pass TERMINAL API, I think the method should only return terminal api because that basically is/are our devices? I can't think of a situation where we might want non terminal api devices (at least retrieval - we might want ability to create a device code for POS sign on later for puck)

JElgar: I think the point about `TERMINAL API` is valid, however, I think it's likely in the future we have greater integration with square devices so returning unpaired and nonterminal devices will become something we want to do. I've updated the filters to filter based on the payment method as this will already be required by the front end to make the pos payment. This should handle the setting of `TERMINAL_API`. You then have a boolean `paired.

atoff (approved):

PR #232, opened 2021-08-17
Feature/booking email

### LINKED ISSUE #39: Create ticket email
- Email user with booking info
- Will need a link to the website where they can view the booking
- Will need a pdf with all the tickets attached maybe

JElgar:
closes #39


uobtheatre/bookings/models.py
```diff
@@ -507,10 +510,41 @@ def pay(self, payment_method: "PaymentMethod") -> Optional["Payment"]:
...
+            "booking": self,
+            "user_name": self.user.first_name.capitalize(),
+            "production_name": self.performance.production.name,
+            "start": self.performance.start,
+            "protocol": "https",
+            "domain": site.domain,
+        }
+
+        subject, from_email, to_email = (
+            "Your booking is confirmed!",
```
atoff, line 538: How about something fancy like "You're going to <production name>!"

uobtheatre/bookings/templates/emails/booking_confirmation_email.html
```diff
@@ -0,0 +1,6 @@
+<p>Hello {{ user_name }}!</p>
+
+<p> Your booking for {{ production_name }} on {{ start|date:"l, d F Y" }} at {{ start|time:"H:i" }} is confirmed!</p>
```
atoff, line 3: Could we make this a bit nicer? Some graphics or something?

uobtheatre/bookings/templates/emails/booking_confirmation_email.txt
```diff
@@ -0,0 +1,6 @@
+Hello {{ user_name }}!
+
```
atoff, line 2: Surely we can get the plain text content from just stripping the HTML tags? Seems very weird to have to write this twice

atoff (changes requested):

uobtheatre/bookings/templates/emails/booking_confirmation_email.txt
JElgar: This assumes the HTML and text versions are the same. I hope someday when we write a proper email the HTML template will deviate maybe with some nice images etc

uobtheatre/bookings/templates/emails/booking_confirmation_email.html
JElgar: I think that's out of the scope of this ticket. This is purely function we should absolutely make a better email with some nice images, better copy and maybe even with the ticket emebedded. I suggest we wait for designs rather than wasting taking making a nice email we know we should throw away

JElgar:
Issue raised for styling https://github.com/BristolSTA/uobtheatre-api/issues/233

PR #236, opened 2021-08-18
feat: Reporting

### LINKED ISSUE #165: Export payment objects to a csv
- GQL mutation, takes date range (start, end)
- Generate CSV file
- Save to S3
- Returns link to pdf

### LINKED ISSUE #246: Production Sales Breakdown (and breakdown permissions)
Expanding on the sales breakdown started in the reports branch the following should be available on a production:
- Total number of tickets sold across performances
- Sum of the total capacity across performances (i.e. the total capacity of the production)
- Sum the number of bookings for a performance
- Have a breakdown of sales by concession type for a performance
- Query the total sales income for the production (this is, the sum of booking payments - positive and negative - deducting any misc costs). This should also be available on a performance basis on a performance node. Must be restricted to authorised users

We should also check that these sensitive sales numbers (such as total capacities and the like) are only available to people with adequate permissions. We don't want regular end users to be able to work out detailed sales information if they inspect our queries / create their own queries.

atoff:
Todo:
- [x]  points from #246
- [x] few TOD!O s regarding permissions to see sales
- [X] implement permissions
- [X] test views
- outstanding payments report must only consider card payments

Closes #246, closes #165


uobtheatre/productions/models.py
```diff
@@ -232,6 +232,31 @@ def duration(self) -> Optional[datetime.timedelta]:
...
+                booking.misc_costs_value() if len(booking.payments.all()) else 0
+                for booking in bookings
+            ]
+        )
+
+        return {
+            "payments_value": total_payments,
+            "misc_costs_value": total_misc_costs,
+            "society_income": total_payments - total_misc_costs,
+        }
```
JElgar, line 258: Lets make these separate properties on the model. If needs be we can make the bookings with payments a cached property to avoid computing that multiple times. Similarly with other properties

uobtheatre/reports/reports.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class MetaItem:
+    def __init__(self, name: str, value):
+        self.name = name
+        self.value = value
+
+
+class DataSet:
+    """A data set represents a list of data items."""
```
JElgar, line 23: Plz use a dataclass for these

uobtheatre/reports/reports.py
```diff
@@ -0,0 +1,130 @@
...
+class DataItem:
+    def __init__(self, data, subject=None):
+        self.subject = subject
+        self.data = data
+
+
+class MetaItem:
+    def __init__(self, name: str, value):
+        self.name = name
+        self.value = value
```
JElgar, line 19: Please explain what these are in comments. I have no idea

uobtheatre/reports/reports.py
```diff
@@ -0,0 +1,130 @@
...
+        return self.find_item_by_subject(subject) or self.create_item(data, subject)
+
+    def create_item(self, *args) -> DataItem:
+        """Creates a data item in the data set"""
+        item = DataItem(*args)
+        self.items.append(item)
+        return item
+
+
+class AbstractReport(ABC):
```
JElgar, line 47:
```suggestion
class Report(ABC):
```

uobtheatre/reports/reports.py
```diff
@@ -0,0 +1,130 @@
...
+
+        # Add all providers
+        for provider in payment_methods.PaymentMethod.__all__:
+            provider_totals_set.create_item(0, provider.name)
+
+        for payment in payments:
+            if payment.pay_object and payment.pay_object.performance:
+
+                # Handle production
+                production_set = production_totals_set.find_or_create_item_by_subject(
```
JElgar, line 78:
```suggestion
                production_item = production_totals_set.find_or_create_item_by_subject(
```

uobtheatre/reports/schema.py
```diff
@@ -0,0 +1,55 @@
...
+from uobtheatre.utils.schema import AuthRequiredMixin
+
+available_reports = {
+    "PeriodTotals": {
+        "cls": reports.PeriodTotalsBreakdown,
+        "uri": "period_totals",
+    },
+    "OutstandingPayments": {
+        "cls": reports.OutstandingSocietyPayments,
+        "uri": "outstanding_society_payments",
```
JElgar, line 17: We should alow the api to change the base url of a repot (incase we save to s3 in the future). We could do this by adding "REPORT_PATH" to settings?

uobtheatre/reports/views.py
```diff
@@ -0,0 +1,177 @@
+from django.http import HttpResponse
+from xlsxwriter.utility import xl_rowcol_to_cell
+
+from uobtheatre.reports.utils import ExcelReport
+
+from . import reports
+
+
+def period_totals(request, start_time, end_time):
```
JElgar, line 9: I think these views are doing too much. We should move this logic somewhere else

JElgar:
We also need to think about permissions on this. We should add a "Finance report generation" permission to allow the use of these endpoints.

uobtheatre/reports/schema.py
atoff: These currently return paths (like /reports/outstanding_payments), instead of fully qualified. I think for now, best way is for the front end to say: if we have just a path, use the API as the base, else (if fully qualified url) we'll use that. Gives us scope to use s3 in future

uobtheatre/productions/schema.py
```diff
@@ -150,6 +163,18 @@ def resolve_is_bookable(self, info):
     def resolve_min_seat_price(self, info):
         return self.min_seat_price()
 
+    def resolve_sales_breakdown(self, info):
+        if not info.context.user.has_perm("productions.sales", self):
```
JElgar, line 167: Is this bit tested?

PR #249, opened 2021-08-28
Feature/society members

### LINKED ISSUE #203: Society Members and Box Office Permissions
Need to implement society members, with the ability/permissions to box office that society's performances.

Need to have a query too, so that we can get the societies that a user is a member of

This is going to be required for MVP/launch

JElgar:
- Societies now have members
- Added abilities (computed permissions)

fixes #203


schema.graphql
```diff
@@ -594,16 +597,6 @@ type UserNode implements Node {
...
-type UserNodeConnection {
-  pageInfo: PageInfo!
-  edges: [UserNodeEdge]!
-}
-
-type UserNodeEdge {
-  node: UserNode
-  cursor: String!
-}
-
```
JElgar, line 606: We should add this back in properly in a bit. Would be really handy if we could work out how to change the UserNode for this package.

uobtheatre/payments/migrations/0003_alter_payment_provider.py
```diff
@@ -0,0 +1,26 @@
...
+            model_name="payment",
+            name="provider",
+            field=models.CharField(
+                choices=[
+                    ("CASH", "CASH"),
+                    ("CARD", "CARD"),
+                    ("SQUARE_POS", "SQUARE_POS"),
+                    ("SQUARE_ONLINE", "SQUARE_ONLINE"),
+                ],
+                max_length=20,
```
JElgar, line 23: Whoopsies, this should already be in main

uobtheatre/users/models.py
```diff
@@ -51,3 +56,29 @@ def has_perm(self, perm: str, obj=None) -> bool:
...
+        Args:
+            permissions: (Union[str, list[str]]): The permissions being
+                checked. If a list is provided only one of the permissions must
+                be met to return True
+
+        Returns:
+            bool: Whether the user has any of the permissions
+        """
+        return (
+            # Here we check explicitly check globalal permissions as no objects
```
atoff, line 80:
```suggestion
            # Here we check explicitly check global permissions as no objects
```

atoff (commented):
Just to make sure I have got this right: a user belonging to a society doesn't give them any special treatment? They need to have a `production.create_production` permission (assigned to that society??) to be able to access admin panel and create a production on behalf of the society?

Also, is it out of scope for this to have an query (and mutation to add/remove) to return users who have boxoffice permission/ability for a given production? 

Is this closing #203?

JElgar:
Yes a user being part of a society does not give them special treatment. However, I think if we are using the front end to allow someone to give permissions, it might be nice to allow them to select from a list of society members.

Yeah that is out of scope but we need to decide what endpoints we need, there will probably be quite a few 😬 

How do you feel about something like this: when a society is created we create a load of groups for them. Eg if MTB is created we then create:
- MTB admin - can view and edit all production as well as edit society information. Also assingn other members the "society groups"/ permissions
- MTB production team - can create show listings (edit perms can be added manually as well)
- MTB viewer - can view productions but not edit (edit perms can be added manually as well)
- MTB finance - can do/see money things?
and we can link those groups directly to the society so we know which groups are society specific. The president of the society can then be given admin etc

These example groups are a bit random so take that with a pinch of salt.

I think giving the creator of a production read and write access is also a very good shout, we might want to also allow someone to add users though when they create the listing (from a list of society members)?

atoff:
Yeah that all sounds quite reasonable, apart from maybe the group's thing, not sure how I feel about flooding the db with loads of groups per society, seems a bit whack imo.

 I also think for the time being, we could just allow only the production creator to have perms to do things with the production until we have a good solution ready for adding members into the abilities as time do be ticking now

JElgar:
closes #203

streth11 (approved):

PR #258, opened 2021-09-04
Fixes/g2g issues

### LINKED ISSUE #235: Society slug not found error
When calling the society query with an unknown/invalid slug, instead of just returning null, an exception is thrown

### LINKED ISSUE #245: Performance Total Capacity
Ref: https://github.com/BristolSTA/uobtheatre-api/blob/main/uobtheatre/productions/models.py#L421

The total capacity function on the performance model is a bit misleading - it is only the sum of the capacity of the seat groups.
It does not take into account the capacity of the performance (if it is set). 

This might be by design, but the name of total is misleading, it should be more like `seat_group_capacity` or similar. 

This function is currently used in the calculation of the "total capacity" in the tickets breakdown, which is therefor inaccurate if we set the capacity of the performance?

In this case, we should expose a calculated property on the model which is the actual "total_capacity" (i.e. the Minimum from the sum of seat group capacities, and the performance capacity if defined). 

Example:
Performance Capacity: 100
Sum of seat groups capacity: 150

Total Capacity Output would be 150, when it should be 100
Remaining Capacity would be 100 (correct)

Example 2:
Performance Capacity: 100
Sum of seat groups capacity: 150
Sold tickets/reduction is seat group capacity: 50

Total Capacity Output would be 150, when it should be 100
Remaining Capacity would be 50 (correct)

Example from GraphQLi:
image

Recommended Changes:
- Rename existing `total_capacity` function to `seat_group_capacity`
- Create new `total_capacity` function to correctly calculate the capacity of the performance factoring in seat groups and performance capacity
- Create a `totalCapacity` property on the `PerformanceNode`

### LINKED ISSUE #256: Misc costs not 0 for comp booking
On a comp booking (admin discount = 1), the misc cost is able to be non-zero:
image

streth11:
closes #256 
closes #245
closes #235

### DIFF
--- a/uobtheatre/productions/models.py
+++ b/uobtheatre/productions/models.py
@@ -419,11 +419,8 @@ class Performance(TimeStampedMixin, models.Model):
             .exists()
         )
 
-    def total_capacity(self, seat_group=None):
-        """Total capacity of the Performance.
-
-        The is the total number of seats (Tickets) which can be booked for this
-        performance. This does not take into account any existing Bookings.
+    def total_seat_group_capacity(self, seat_group=None):
+        """The sum of the capacities of all the seat groups in this performance.
 
         Note:
             The sum of the capacities of all the seat groups is not necessarily
@@ -434,7 +431,7 @@ class Performance(TimeStampedMixin, models.Model):
                 (default None)
 
         Returns:
-            int: The capacity of the show (or SeatGroup if provided)
+            int: The capacity of the seat groups (or SeatGroup if provided) of the show
         """
         if seat_group:
             queryset = self.performance_seat_groups
@@ -444,6 +441,21 @@ class Performance(TimeStampedMixin, models.Model):
                 return 0
         response = self.performance_seat_groups.aggregate(Sum("capacity"))
         return response["capacity__sum"] or 0
+
+    def total_capacity(self, seat_group=None):
+        """Total capacity of the Performance.
+
+        The is the total number of seats (Tickets) which can be booked for this
+        performance. This does not take into account any existing Bookings.
+
+        Returns:
+            int: The capacity of the show
+        """
+        if seat_group is not None:
+            return self.total_seat_group_capacity(seat_group=seat_group)
+        if self.capacity:
+            return min(self.capacity, self.total_seat_group_capacity())
+        return self.total_seat_group_capacity()
 
     def total_tickets_sold(self, **kwargs):
         """The number of tickets sold for the performance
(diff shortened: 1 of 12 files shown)


uobtheatre/productions/models.py
```diff
@@ -445,6 +442,21 @@ def total_capacity(self, seat_group=None):
...
+
+        Returns:
+            int: The capacity of the show
+        """
+        if seat_group is not None:
+            return self.total_seat_group_capacity(seat_group=seat_group)
+        if self.capacity:
+            return min(self.capacity, self.total_seat_group_capacity())
+        return self.total_seat_group_capacity()
+
```
streth11, line 459: @JElgar I dont know if this is the best way of solving the issue (#245) but it seems to work

uobtheatre/productions/models.py
```diff
@@ -445,6 +442,21 @@ def total_capacity(self, seat_group=None):
         response = self.performance_seat_groups.aggregate(Sum("capacity"))
         return response["capacity__sum"] or 0
 
+    def total_capacity(self, seat_group=None):
```
atoff, line 445: This should be a propety probably, and it as such it probably shouldn't accept a seat group argument. There are places where this was used - this will need to be refactored to use the new seat group capacity func

uobtheatre/productions/models.py
```diff
@@ -445,6 +442,21 @@ def total_capacity(self, seat_group=None):
...
+        The is the total number of seats (Tickets) which can be booked for this
+        performance. This does not take into account any existing Bookings.
+
+        Returns:
+            int: The capacity of the show
+        """
+        if seat_group is not None:
+            return self.total_seat_group_capacity(seat_group=seat_group)
+        if self.capacity:
+            return min(self.capacity, self.total_seat_group_capacity())
```
atoff, line 457: Could refactor this into a one liner
`return min(...) if self.capacity else ...`

atoff (changes requested):

uobtheatre/productions/models.py
JElgar: I'm confused, probably just being slow but we should chat about this

uobtheatre/productions/models.py
streth11: image

i prefer the left, but will commit the right if you lot prefer

atoff: I still prefer right but its not a deal breaker

atoff (approved):

PR #265, opened 2021-09-21
fix: Max tickets per bookings, and booking reservation expiry

JElgar:


atoff:
@JElgar can you review this and check you are happy when you get a minute?

JElgar (commented):
This looks good but I think I might need someone to chat me through how we've handled capacity remaining because I'm a bit lost

JElgar:
Approved

PR #269, opened 2021-09-22
feat: Better Emails

### LINKED ISSUE #233: Beautify emails
Currently, the emails are plain text. We should create a css file/css files to style the emails as well as adding some nice images and updating the copy.

Email clients do not support external CSS but we can get around that using a package like this:
https://github.com/roverdotcom/django-inlinecss

This should allow us to create a base css file which we can use for all emails.

atoff:
Make emails great again!

Closes #233
image


uobtheatre/mail/composer.py
```diff
@@ -0,0 +1,155 @@
+import abc
```
JElgar, line 1: This is massively overcomplicating a very simple task. Please don't do this. Just write 2 templates. We won't want the same format in the plain text email anyway.

JElgar (changes requested):
You seem to really be getting the hang of how to use email templates. Please keep that up and remove this horrifying email composed. Much love.

config/settings/common.py
```diff
@@ -70,6 +71,11 @@
 
 # Email
 EMAIL_BACKEND = "django.core.mail.backends.smtp.EmailBackend"
+EMAIL_HOST = os.getenv("EMAIL_HOST", "localhost")
+EMAIL_PORT = os.getenv("EMAIL_PORT") or 1025
+EMAIL_HOST_USER = os.getenv("EMAIL_HOST_USER")
+EMAIL_HOST_PASSWORD = os.getenv("EMAIL_HOST_PASSWORD")
+DEFAULT_FROM_EMAIL = "UOB Theatre <no-reply@uobtheatre.com>"
```
JElgar, line 78: Can this be moved into `local.py`? (Since we have a different setup for production)

JElgar (changes requested):
I think we should also add some unit tests to MailComposer

config/settings/common.py
atoff: the default backend is defined in each (i.e. local and prod) but the host, port and user/pass stuff is should be defined in common so we can use SMTP instead of SES for example in production. I don't think we want to lock our production into just SES

JElgar (approved):
LGTM

PR #270, opened 2021-09-22
feat: Free bookings and other stuff

### LINKED ISSUE #190: Handle free bookings
When a booking's total comes to 0, payBooking migration should just set the status of the booking to Paid.

atoff:
Allows for free bookings (by not creating a payment/parsing the payment provider)
Also enforces the need for a booking to have at least one ticket to be paid for

Closes #190

### DIFF
--- a/uobtheatre/bookings/schema.py
+++ b/uobtheatre/bookings/schema.py
@@ -634,34 +634,43 @@ class PayBooking(AuthRequiredMixin, SafeMutation):
         if booking.status != Booking.BookingStatus.IN_PROGRESS:
             raise GQLException(message="The booking is not in progress")
 
-        if payment_provider == SquareOnline.name:
-            if not nonce:
-                raise GQLException(
-                    message=f"A nonce is required when using {payment_provider} provider.",
-                    field="nonce",
-                    code="missing_required",
-                )
-            payment_method = SquareOnline(nonce, booking.id)
+        # Booking must have at least one ticket
+        if booking.tickets.count() == 0:
+            raise GQLException(message="The booking must have at least one ticket")
 
-        elif payment_provider == SquarePOS.name:
-            if not device_id:
-                raise GQLException(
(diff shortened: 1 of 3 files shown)


uobtheatre/bookings/schema.py
```diff
@@ -634,34 +634,43 @@ def resolve_mutation(  # pylint: disable=too-many-arguments
...
-                raise GQLException(
-                    message=f"A device_id is required when using {payment_provider} provider.",
-                    field="device_id",
-                    code="missing_required",
+        # Booking must have at least one ticket
+        if booking.tickets.count() == 0:
+            raise GQLException(message="The booking must have at least one ticket")
+
+        # If the booking isn't free, we care about the payment provider. Otherwise, we don't
+        if booking.total() > 0:
```
JElgar, line 642: Can we not nest this? So if just the first branch in the `elif`?
```suggestion
        if booking.total() == 0:
            # Free bookings do not require a payment
            pass
```

uobtheatre/bookings/test/test_mutations.py
```diff
@@ -21,6 +22,19 @@
 from uobtheatre.venues.test.factories import SeatFactory, SeatGroupFactory
 
 
+def add_ticket_to_booking(booking):
```
JElgar, line 25: Can we move this into bookings/test/factories

JElgar: Maybe with some (optional?) arguments?

JElgar (changes requested):

uobtheatre/bookings/test/test_mutations.py
atoff: if we need to extend later with arguments we can, I think for now this should be enough, its mainly just to make it a valid booking

JElgar (approved):
LGTM

PR #275, opened 2021-09-25
fix: Square fixes

atoff:
- Make square error messages from the API more informative
- Shift responsibility for supplying the idempotency key to frontend


JElgar (approved):
Why have we moved this? Is this because you now think it's the right place for it or is this just a quick fix?

JElgar (approved):

PR #278, opened 2021-09-29
Added processing fee and app fee to payments

JElgar:


local.yml
```diff
@@ -17,7 +17,6 @@ services:
       - .:/app/uobtheatre-api:z
     env_file:
       - ./.envs/.local/.django
-      - .env
```
atoff, line 20: Can we leave this in. It provides a good way to override the environment defined in the tracked files on a local deployment.

uobtheatre/payments/payables.py
```diff
@@ -8,10 +12,41 @@ class Payable(metaclass=AbstractModelMeta):
...
+        ]
+
+    @property
+    def society_payment_value(self) -> int:
+        """The amount owed to the society for selling this object."""
+        return self.payments.filter(type=Payment.PaymentType.PURCHASE).aggregate(  # type: ignore
+            value=Sum("value") - Sum("app_fee")
+        )[
+            "value"
+        ]
```
atoff, line 52: There is no logic in this file to calculate the amount to transfer to the society. (Description on line 47 is a bit incorrect in this sense)

I think we need another property like `society_transfer_value` or something which is basically `society_payment_value` but filtering for non-cash transactions? Also, what if `app_payment_value` comes out negative? I.e. our fee minus provider fee is negative? This difference would need to be deducted from a `society_transfer_value` in that instance.

We likely also want a property to calculate the total payments value (i.e. sum the value (total charged to customer)).

From the report point of view, the numbers we are interested for a production are: Total Sales, Total increase in square balance, Total fees for us, amount to be transferred to society.

Another point - how does this scale? On a production which might have over 300/400 payments, this looks like a lot of queries if we loop every performance, every booking and every payment.

atoff (changes requested):
Hopefully some of these comments make sense

local.yml
JElgar: If a `.env` file is not found an error is thrown. If we want to add local .envs they should be shoved in .envs/.local

atoff: having an .env in the project root to override the "typical" environment variables is standard practice (and we need to to add, for example, square details and change mail backends, etc), it makes it easy and clear for people rather than having to go into some complicated folder structure. Either way, can you implement some way that isn't tracked or something, instead of just removing it and leaving it

uobtheatre/payments/payables.py
JElgar: For now I would say:
- Performance of reports is a future problem
- If `app_payment_value` thats bad but for now lets just make sure that doesnt happen (we currently charge 5% theatre improvement levy and square charges us 1.9% so we should be good)

atoff (approved):

PR #282, opened 2021-10-02
Remove can-box-office

streth11:


config/settings/test.py
```diff
@@ -8,6 +8,7 @@
 # ------------------------------------------------------------------------------
 # https://docs.djangoproject.com/en/dev/ref/settings/#secret-key
 TEST_RUNNER = "django.test.runner.DiscoverRunner"
+BASE_URL = "https://example.com"
```
JElgar, line 11: Can we change this so its not example.com. This is because the default site for testing is also example.com
```suggestion
BASE_URL = "https://api.example.com"
```

JElgar (approved):
LGTM

PR #283, opened 2021-10-02
fix: Draft Booking Deletion with Creator

atoff:
Uses the target user of the booking instead of the current auth user


JElgar (approved):
LGTM

PR #298, opened 2021-10-06
fix: Update booking target user

atoff:
When updating the target user on an existing booking, if that user already has a draft booking for that performance, an error gets thrown. This will delete the previous booking

### DIFF
--- a/uobtheatre/bookings/test/test_mutations.py
+++ b/uobtheatre/bookings/test/test_mutations.py
@@ -948,7 +948,6 @@ def test_update_booking_no_tickets(gql_client):
 
 @pytest.mark.django_db
 def test_update_booking_set_target_user(gql_client):
-
     creator = gql_client.login().user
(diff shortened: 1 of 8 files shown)


uobtheatre/bookings/test/test_mutations.py
```diff
@@ -948,14 +948,19 @@ def test_update_booking_no_tickets(gql_client):
...
     creator.is_superuser = True
     creator.save()
 
     user = UserFactory(email="user@email.com")
+    # Give the user and existing draft booking
+    BookingFactory(
+        user=user,
+        performance=booking.performance,
+        status=Booking.BookingStatus.IN_PROGRESS,
+    )
```
JElgar, line 963: Should this booking be deleted? If so can be add an assert for this to the test

JElgar (approved):
LGTM

PR #299, opened 2021-10-06
Handle square response with multiple payments

JElgar:


uobtheatre/payments/test/test_payment_methods.py
```diff
@@ -168,13 +151,30 @@ def test_square_pos_pay_success(mock_square):
...
+                },
+                "status": "PENDING",
+            }
+        },
     ):
-        payment_method = SquarePOS("device_id")
+        payment_method = SquarePOS("device_id", "ikey")
         payment_method.pay(100, 14, BookingFactory())
 
     # Assert no payments are created. A payment should only be created by the
```
atoff, line 168: Old comment

uobtheatre/societies/test/test_schema.py
```diff
@@ -7,9 +7,9 @@
 
 @pytest.mark.django_db
 def test_societies_schema(gql_client):
-    societies = [SocietyFactory() for i in range(3)]
+    societies = [SocietyFactory() for _ in range(3)]
     society_productions = [
-        [ProductionFactory(society=society) for i in range(10)] for society in societies
+        [ProductionFactory(society=society) for _ in range(1)] for society in societies
```
atoff, line 12: Any particular reason for this?

atoff (changes requested):
If we receive another "payBooking" mutation for a booking with an existing "In progress" payment, we should cancel that existing in progress booking.

e.g. If we send `payBooking` for POS, and then send `payBooking` with POS again, the previous payment should be cancelled (w/ square) and then deleted from our system too, and a new payment created. (There might be an edge case here where we haven't received the webhook for cancellation/payment success from square, in which case we should check whilst doing the cancel at square)

Equally, if we send `payBooking` for POS, and then send `payBooking` with CASH, the inprogress square pos should be  cancelled (w/ square) and then deleted from our system too.

At the moment it is possible to create multiple payments on a booking this way

uobtheatre/payments/payment_methods.py
```diff
@@ -207,8 +256,61 @@ def list_devices(cls, product_type: str = None, status: str = None) -> list[dict
...
-class SquareOnline(PaymentMethod):
+        Returns:
+            int: The processing fee
+        """
+        checkout = cls.get_checkout(checkout_id)
+        processing_fees = [
+            cls.payment_processing_fee(cls.get_payment(payment_id))
+            for payment_id in checkout["payment_ids"]
+        ]
+        return sum(filter(None, processing_fees))
```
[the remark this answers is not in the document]
JElgar: Yes

atoff (approved):

PR #308, opened 2021-10-08
Prod db

streth11:


Makefile
```diff
@@ -41,6 +41,9 @@ down: ## Down
 dump: ## dumps databse objects into fixture
 	docker-compose run --rm api python manage.py dumpdata users images addresses venues societies productions discounts bookings payments --indent 2 > db.json
 
+dump-prod: ## dumps databse objects into fixture
+	docker-compose run --rm api python manage.py dumpdata users images addresses venues societies productions discounts bookings payments --indent 2 > db-prod.json
+
```
JElgar, line 46: We shouldn't need this again so Id say lets just remove this
```suggestion
```

Makefile
```diff
@@ -62,6 +65,9 @@ seed: ## Seed the db with some example data
 seed-testfixtures: ## Seed the data for e2e testing
 	docker-compose run api python manage.py loaddata db.json
 
+seed-prod: ## Seed the data for e2e testing
+	docker-compose run api python manage.py loaddata db-prod.json
+
```
JElgar, line 70:
```suggestion
```

Same as above

Makefile
```diff
@@ -136,3 +142,6 @@ build:
 
 django-shell: ## Open django shell
 	docker-compose run --rm api python manage.py shell
+
+flush:
+	docker-compose run api python manage.py flush
```
JElgar, line 147:
```suggestion
flush:
	docker-compose run --rm api python manage.py flush
```

db-prod.json
```diff
@@ -0,0 +1,694 @@
+[
```
JElgar, line 1: I was gonna say let's not add this, but to make life extra easy let's just merge this in and we can delete later/ add the data to the other seeded data

PR #321, opened 2021-10-12
fix: Performance and Production Query Visibilities

### LINKED ISSUE #320: Production "User Can See" on single node
People are able to view the productions in draft, even if they don't have any of the associated admin permissions
They should only be able to see productions that are not in draft/submitted (i.e. not publish)

atoff:
Closes #320, but also implements it on the performance model

NB: Need to do same for the booking


JElgar (approved):
LGTM

PR #331, opened 2021-10-17
Admin qol improvements

JElgar:


JElgar:
@atoff what else would be handy?

PR #343, opened 2021-10-31
feat: Email Ticket Links

### LINKED ISSUE #274: Venue timezones
We should implement venue timezones (or be able to find the timezone), so that emails show booking times in the correct local time

atoff:
Based on new frontend features implemented in https://github.com/BristolSTA/uobtheatre-web/pull/214, we can now generate links to allow users to view their tickets without being logged in.

This PR:
- Generates this link and inserts it into the booking email
- Adds methods to guess the event's timezone based on its venue's address. This timezone is then used in the email generator for the booking email to give the user localised times and reduce confusion

Closes #274

### DIFF
--- a/uobtheatre/bookings/models.py
+++ b/uobtheatre/bookings/models.py
@@ -588,13 +588,14 @@ class Booking(TimeStampedMixin, Payable, models.Model):
 
         composer.line(
             (
-                "This event opens at %s for a %s start. Please bring your tickets (printed or on your phone) or your booking reference (<strong>%s</strong>)."
+                "This event opens at %s for a %s start (please note that %s might not be your current timezone). Please bring your tickets (printed or on your phone) or your booking reference (<strong>%s</strong>)."
                 if self.user.status.verified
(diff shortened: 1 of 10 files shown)


JElgar (approved):
LGTM

uobtheatre/bookings/models.py
```diff
@@ -554,6 +555,20 @@ def complete(self):
...
+        query_string = "?" + "&".join(
+            [
+                f"performanceID={to_global_id('PerformanceNode', self.performance.id)}",
+            ]
+            + [
+                f"ticketID={to_global_id('TicketNode', id)}"
+                for id in self.tickets.values_list("id", flat=True)
+            ]
+        )
+        return f"/user/booking/{self.reference}/tickets" + query_string
```
JElgar, line 570: We should use `urlencode(params, True)` here (from urllib.parse import urlencode)

JElgar (changes requested):

JElgar (approved):
LGTM

PR #352, opened 2021-11-09
fix: Unique discount validation

atoff:


uobtheatre/discounts/models.py
```diff
@@ -90,6 +90,7 @@ def validate_unique(self, *args, **kwargs):
             discount
             for discount in discounts
             if discount.get_concession_map() == self.get_concession_map()
+            and discount.percentage == self.percentage
```
JElgar, line 93: This should be a performance

JElgar (changes requested):

JElgar (approved):
LGTM

PR #354, opened 2021-11-12
feat: Production admin filtering

### LINKED ISSUE #316: Node Filtering by Permissions
The main problem at the moment is being able to filter productions by productions on which the user can administrate, but we could apply this more generally with some sort of filter mixin thing so that:

- On each model with object-level permissions, we should be able to provide an input filter with a list of permissions for either an OR match or an ALL match (i.e. the user has one of the given permissions, or all of the given permissions)

e.g. `productions(userHasSomePermissions: ["permission1"])` and `productions(userHasPermissions: ["permission1", "permission2"])`

atoff:
Closes #316


JElgar (approved):
Nice!

PR #372, opened 2021-12-02
feat: Booking queries outside user's domain

atoff:


JElgar (approved):
Should we do the same for PaymentNode?

atoff:
> Should we do the same for PaymentNode?

note sure how we'd do this because it is a polymorphic relation. I'm going to merge this for now as it is quite time critical

PR #397, opened 2022-01-06
fix: Performance Capacity Calculation

### LINKED ISSUE #393: Venue with multiple seatgroups doesnt cap out to venue capacity
Sum of seatgroups is over 207
image

Performacne capacity is sum of seatgroups not capped at venue
image

Query
image

atoff:
Fix the current calculation by factoring in the internal capacity on a venue as a limiting factor on the number of tickets able to be sold at a venue

Fixes #393

### DIFF
--- a/uobtheatre/productions/models.py
+++ b/uobtheatre/productions/models.py
@@ -332,11 +332,15 @@ class Performance(
             int: The capacity of the show
         """
 
-        return (
-            min(self.capacity, self.total_seat_group_capacity())
-            if self.capacity
-            else self.total_seat_group_capacity()
-        )
+        limiting_capacities = [
+            self.total_seat_group_capacity(),
+            self.venue.internal_capacity,
+        ]
+
+        if self.capacity:
+            limiting_capacities.append(self.capacity)
+
+        return min(limiting_capacities)
 
     def total_tickets_sold(self, **kwargs):
         """The number of tickets sold for the performance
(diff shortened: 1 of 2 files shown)


JElgar (commented):
I think we also need to factor venue capacity into the 2nd branch of `capacity_remaining`

streth11:
on booking 1 ticket in 'the meh seats' the seat group capacity on `generic seating` remains at 207 instaid of 206. Surely the seat group capacity remaining cannot be larger than the venue capacity remaining.

image

atoff:
> on booking 1 ticket in 'the meh seats' the seat group capacity on `generic seating` remains at 207 instaid of 206. Surely the seat group capacity remaining cannot be larger than the venue capacity remaining.
> 
> image

Currently, we treat seat groups as sovereign - i.e. their remaining capacity only goes down when a ticket is sold in that group. This is a good representation because when asking about the seat group capacity, we don't want the performance to influence.
We account for the actual capacity of the performance later in the checks. (See https://github.com/BristolSTA/uobtheatre-api/blob/dev/uobtheatre/productions/models.py#L533)

JElgar:
@streth11 I think that's a great point. Personally, I think we should take the min of the performance capacity remaining and the seat group capacity remaining for that value. I don't think this is logic we should handle on the front end (which is how we do it now). @streth11 would you mind making a system test out of the thing you just wrote. Basically, slap that query into a file called system_test in productions/test and create the required fixtures. Then make the response as you would expect eg with the capacity of the seatgroups limited by the total capacity remaining

PR #417, opened 2022-01-17
fix: Booking creation policy

### LINKED ISSUE #311: General Users can create a booking for a draft produciton
Performances are regarded as bookable even if the the status isn't published.

A performance should not be bookable when the status isn't published, with the exception of society staff with permission to edit the production (change production). In this case, they should be able to create comp bookings in all status' except closed or completed.

They probably also shouldn't be able to query drafts that they cant see either...

atoff:
Closes #311

### DIFF
--- a/uobtheatre/productions/abilities.py
+++ b/uobtheatre/productions/abilities.py
@@ -73,7 +73,6 @@ class BookForPerformance(Ability):
                 and obj.production.status
                 in [
                     Production.Status.APPROVED,
-                    Production.Status.PUBLISHED,
                 ]  # and the status is approved or published (e.g. creating comps)
             )
         )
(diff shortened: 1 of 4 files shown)


uobtheatre/productions/abilities.py
```diff
@@ -46,3 +47,33 @@ def user_has_for(user, obj) -> bool:
...
+            == Production.Status.PUBLISHED  # Must be bookable and published
+            or (
+                (
+                    user.has_perm("productions.change_production", obj)
+                    or user.has_perm("productions.force_change_production", obj)
+                )  # If the user can edit the production
+                and obj.production.status
+                in [
+                    Production.Status.APPROVED,
+                    Production.Status.PUBLISHED,
```
JElgar, line 76:
```suggestion
```

JElgar (approved):

PR #420, opened 2022-01-19
fix: Raise exception for payments that cant be refunded

atoff:


JElgar (approved):
We will get some conflicts but we can resolve later

PR #452, opened 2022-02-16
fix: Cancelled Terminal Webhooks

atoff:


uobtheatre/payments/transaction_providers.py
```diff
@@ -587,6 +587,10 @@ def pay(
             currency=amount_details["currency"],
         )
 
+    @classmethod
+    def cancel(cls, payment: "payment_models.Transaction") -> None:
+        payment.delete()
```
JElgar, line 592: Should we have this? I don't think a square online payment can ever be canceled?

JElgar (changes requested):

PR #479, opened 2022-03-12
Feature/cloudwatch

JElgar:


Makefile
```diff
@@ -158,5 +158,8 @@ build:
 django-shell: ## Open django shell
 	$(COMMAND_PREFIX) python manage.py shell
 
+attach: ## Open django shell
```
JElgar, line 161:
```suggestion
attach: ## Attach to api instance
```

config/settings/production.py
```diff
@@ -59,6 +61,33 @@
     "Cache-Control": "max-age=86400, s-maxage=86400, must-revalidate",
 }
 
+# Cloudwath logging
```
atoff, line 64:
```suggestion
# Cloudwatch logging
```

local.yml
```diff
@@ -21,6 +21,8 @@ services:
     ports:
       - '8000:8000'
     command: /start
+    stdin_open: true
+    tty: true
```
atoff, line 25: I assume this is to support the attach command you added? Is this only for testing on local?

config/settings/production.py
```diff
@@ -59,6 +61,33 @@
...
+    "aws": {
+        "format": "%(asctime)s [%(levelname)-8s] %(message)s [%(pathname)s:%(lineno)d]",
+        "datefmt": "%Y-%m-%d %H:%M:%S",
+    },
+}
+LOGGING["handlers"]["cloudwatch"] = {  # type: ignore
+    "level": "INFO",
+    "class": "watchtower.CloudWatchLogHandler",
+    "boto3_session": logger_boto3_session,
+    "log_group": env("CLOUDWATCH_LOG_GROUP_NAME"),
```
atoff, line 79: Where is `CLOUDWATCH_LOG_GROUP_NAME` getting implemented?

atoff (commented):
Have you managed to test this locally?

local.yml
JElgar: Handy for debugging as you can add a break and get into pdb

PR #485, opened 2022-03-22
feat: Query venues on a production

atoff:
Implements ability to query the venues associated with a production from a production model/node


JElgar (commented):
Why do we have both venues and unique_venues?

atoff:
> Why do we have both venues and unique_venues?

venues = relationship. As it's many-to-many, if there are three performances all with Theatre A, it will return 3 lots of Theatre A
unique_venues = property, just with the distinct operation pre-applied

JElgar (approved):

PR #493, opened 2022-04-02
feat: Report content via GraphQL

atoff:
Expose report data via GraphQL schema

### DIFF
--- a/uobtheatre/reports/schema.py
+++ b/uobtheatre/reports/schema.py
@@ -62,7 +62,6 @@ class GenerateReport(AuthRequiredMixin, SafeMutation):
     report = graphene.Field(ReportNode)
 
     def resolve_report(self, _):
-        print("HERE")
         self.report.run()
         return self.report
 
--- a/uobtheatre/reports/schema.py
+++ b/uobtheatre/reports/schema.py
@@ -61,6 +61,11 @@ class GenerateReport(AuthRequiredMixin, SafeMutation):
     download_uri = graphene.String()
     report = graphene.Field(ReportNode)
 
+    def resolve_report(self, _):
+        print("HERE")
+        self.report.run()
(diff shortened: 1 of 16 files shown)


uobtheatre/reports/schema.py
```diff
@@ -43,6 +59,12 @@ class Arguments:
         options = graphene.List(ReportOption)
 
     download_uri = graphene.String()
+    report = graphene.Field(ReportNode)
+
+    def resolve_report(self, _):
+        print("HERE")
```
JElgar, line 65:
```suggestion
```

uobtheatre/reports/schema.py
```diff
@@ -33,6 +33,22 @@ class ReportOption(graphene.InputObjectType):
...
+    name = graphene.String(required=True)
+    value = graphene.String(required=True)
+
+
+class ReportNode(graphene.ObjectType):
+    datasets = graphene.List(DataSetNode)
+    meta = graphene.List(MetaItemNode)
+
+
 class GenerateReport(AuthRequiredMixin, SafeMutation):
```
JElgar, line 52: Should this not be a query? It feels like this is not mutating any data but just getting something

JElgar (changes requested):

uobtheatre/reports/schema.py
atoff: I think it should stay a mutation because it is doing a (fairly expensive) operation by collating the data. This can include calls to Square, etc. This is much more a complex algorithm compared to a simply property on a model (which is essentially what querying is for) - you are executing a function here that can actually do work and update / change things - the definition of a mutaiton.

atoff:
Can we get this merged @JElgar?

JElgar (approved):
LBTM

PR #514, opened 2022-05-06
feat: Delete In Progress Bookings

atoff:
BREAKING FRONTEND CHANGES

### DIFF
--- a/uobtheatre/bookings/test/test_mutations.py
+++ b/uobtheatre/bookings/test/test_mutations.py
@@ -2226,7 +2226,11 @@ def test_paybooking_unsupported_payment_provider(info):
 
     with pytest.raises(GQLException) as exc:
         PayBooking.resolve_mutation(
-            None, info, booking.id, booking.total, payment_provider="NOT_A_THING"
+            None,
+            info,
+            id=booking.id,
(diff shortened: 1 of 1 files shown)


uobtheatre/bookings/mutations.py
```diff
@@ -165,6 +165,16 @@ def delete_user_drafts(user, performance_id, booking_id=None):
...
 
 
+def authorize_user_for_booking(info, booking):
+    if booking.user.id != info.context.user.id and not info.context.user.has_perm(
+        "productions.boxoffice", booking.performance.production
+    ):
+        raise AuthorizationException(
+            message="You do not have permission to access this booking",
+            field="booking_id",
+        )
```
JElgar, line 175: Should we pop this in an ability?

uobtheatre/bookings/mutations.py
```diff
@@ -357,6 +362,46 @@ def resolve_mutation(
         return UpdateBooking(booking=booking)
 
 
+class DeleteBooking(AuthRequiredMixin, SafeMutation):
```
JElgar, line 365: Is there a way this can use the DeleteForm thing?

PR #515, opened 2022-05-06
feat: Performance intervals

### LINKED ISSUE #419: Performance Interval
Add an interval duration attribute onto performances, to indicate that a run time includes an interval

atoff:
Introduces a new attribute to store performance interval on performance models

TODO: Change this to boolean not integer - realistically we don't care that much about the length, just if there is one or not

Closes #419


JElgar (approved):
LGTM

PR #542, opened 2022-05-30
fix: Interval validation and performance duration

atoff:
- Adds a requirement that intervals < 2 hours long and less than the duration of the performance
- Makes performance duration work across days by using `total_seconds()`


uobtheatre/productions/schema.py
```diff
@@ -337,7 +337,7 @@ def resolve_is_online(self, info):
         return False
 
     def resolve_duration_mins(self, info):
-        return self.duration().seconds // 60
+        return self.duration.total_seconds() // 60
```
JElgar, line 340: 1. Can we make total_seconds a property?
2. Why don't we just make duration_mins a property on the performance then we don't need a custom resolver?

uobtheatre/productions/test/test_schema.py
```diff
@@ -865,9 +873,9 @@ def test_performance_schema(gql_client):
                         "node": {
                             "createdAt": performance.created_at.isoformat(),
                             "updatedAt": performance.updated_at.isoformat(),
-                            "capacity": performance.capacity,
+                            "capacity": None,
```
JElgar, line 876: Why? Does this not depend on the implementation of the factory? (Same for disabled)

JElgar (changes requested):

JElgar (approved):

PR #549, opened 2022-06-12
feat: Better content warnings

### LINKED ISSUE #537: Improved Content Warnings
Currently, content warnings are limited to a fixed set of "warnings" stored in our database. These can be selected and associated with many productions (i.e. a production can have any number of warnings from the available warnings)

From discussions with PAF groups, a new format is required where by:
- There are a fixed list of high-level warning categories (these will be populated in the database from a fixed list in the migration, but addition to these should be possible)
- High level categories can be associated with productions as we do currently, with the change that they now also need to support a free text description of the warning

This will work like so:
Performance A contains references to suicide and depictions of murder. The person creating the performance associates the high-level categories "Themes of suicide" and "Gore" (examples). When associating these content warnings by their IDs, the user should also be able to pass a free-text description (e.g. "References to suicide in Act 2 for less than 30 seconds").

Work Required:

- Update current "warnings" model to a more appropriate name (e.g. "Content Warnings")
- Provide migration strategy to delete existing uses of warnings
- Extend the pivot between Production and ContentWarning models to include a nullable description field
- Create/Update/Modify production mutation process to support associating warnings and descriptions
- Possibly offer a default description on ContentWarning objects

atoff:
Closes #537


atoff:
This is g2g @JElgar

schema.graphql
```diff
@@ -592,14 +611,19 @@ type ProductionTeamMemberNode implements Node {
   production: ProductionNode!
 }
 
+input ProductionWarning {
```
JElgar, line 614: ProductionContentWarningInput

schema.graphql
```diff
@@ -848,8 +872,8 @@ type VenueNode implements Node {
...
-  productions(offset: Int, before: String, after: String, first: Int, last: Int, createdAt: DateTime, updatedAt: DateTime, name: String, subtitle: String, description: String, venues: [ID], society: ID, status: String, ageRating: Int, facebookEvent: String, warnings: [ID], slug: String, id: ID, userHasPermission: String, start: DateTime, start_Gte: DateTime, start_Lte: DateTime, end: DateTime, end_Gte: DateTime, end_Lte: DateTime, search: String, orderBy: String): ProductionNodeConnection
+  productionSet(offset: Int, before: String, after: String, first: Int, last: Int, createdAt: DateTime, updatedAt: DateTime, name: String, subtitle: String, description: String, venues: [ID], society: ID, status: String, ageRating: Int, facebookEvent: String, supportEmail: String, contentWarnings: [ID], slug: String, id: ID, userHasPermission: String, start: DateTime, start_Gte: DateTime, start_Lte: DateTime, end: DateTime, end_Gte: DateTime, end_Lte: DateTime, search: String, orderBy: String): ProductionNodeConnection!
```
JElgar, line 875: What is the thinking with supportEmail? When will the support email not be `support@uobtheatre.com`?

I think if we want people to be able to contact the production teams then we need to consider that separately to support (i.e. maybe add a productionTeamContantEmail). But I wouldn't be super happy just chucking an email on the site so I think this needs some thought. We also don't want to run into a situation where people are emailing the production team about the site being broken....

If the idea is to have different support emails per production then I think this also needs some consideration and probably some integration with os tickets which doesn't exist yet.

Is there a concrete plan of how this is going to be used or is it just always going to be support@ for now. If its always going to be support@ then lets remove it. If not then I think this needs some thought before we merge

uobtheatre/productions/fixtures.json
```diff
@@ -1,146 +1,148 @@
 [
-    {
```
JElgar, line 2: There seem to be a lot of linting changes here. Have you run precommit? I think theres a linter for json in there somewhere... We should make CI fail in future

uobtheatre/productions/forms.py
```diff
@@ -1,10 +1,69 @@
...
+
+class ProductionWarningListField(Field):
+    pass
+
+
+@convert_form_field.register(ProductionWarningListField)
+def convert_form_field_to_string(field):
+    return graphene.List(
+        ProductionWarning, description=field.help_text, required=field.required
+    )
```
JElgar, line 32: I still hate this, can we make this not a thing?

uobtheatre/productions/forms.py
```diff
@@ -1,10 +1,69 @@
...
 
 class ProductionForm(MutationForm):
+    """Form for productions"""
+
+    contentWarnings = ProductionWarningListField(required=False)
+
+    def clean(self):
+        """Validate form data on clean"""
+        cleaned_data = super().clean()
+        warnings = cleaned_data.get("contentWarnings") or []
```
JElgar, line 43: Why does `cleaned_data.get("contentWarnings", [])` not work

uobtheatre/productions/forms.py
```diff
@@ -1,10 +1,69 @@
...
+
+        for warning in warnings:
+            if not ContentWarning.objects.filter(pk=warning["id"]).exists():
+                raise ValidationError(
+                    {
+                        "contentWarnings": f"A warning with ID {warning['id']} does not exist"
+                    }
+                )
+
+    def _save_m2m(self):
```
JElgar, line 53: Can you explain why we have to go with this style of Django forms? I'm not a fan of save_m2m if it can be avoided? I also hate how this is an `_` method?

uobtheatre/productions/forms.py
```diff
@@ -1,10 +1,69 @@
...
+        """Save the many-to-many relations"""
+        super()._save_m2m()
+
+        if (warnings := self.cleaned_data.get("contentWarnings")) is not None:
+            ProductionContentWarning.objects.filter(production=self.instance).delete()
+
+            for warning in warnings:
```
JElgar, line 60: https://docs.djangoproject.com/en/dev/ref/models/querysets/#bulk-create

uobtheatre/productions/migrations/0021_auto_20220612_1317.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.2.13 on 2022-06-12 13:17
+
+from django.db import migrations, models
```
JElgar, line 3: Why are there so many migrations in this pr? Can we squash these?

uobtheatre/productions/models.py
```diff
@@ -859,7 +872,11 @@ def PRIVATE_STATUSES(cls):  # pylint: disable=invalid-name
     age_rating = models.SmallIntegerField(null=True, blank=True)
     facebook_event = models.CharField(max_length=255, null=True, blank=True)
 
-    warnings = models.ManyToManyField(AudienceWarning, blank=True)
+    support_email = models.EmailField()
```
JElgar, line 875: Is this nullable?

uobtheatre/productions/schema.py
```diff
@@ -53,10 +54,16 @@ class Meta:
         interfaces = (relay.Node,)
 
 
+class ProductionContentWarningNode(DjangoObjectType):
+    class Meta:
+        model = ProductionContentWarning
+        interfaces = (relay.Node,)
+
+
 class WarningNode(DjangoObjectType):
```
JElgar, line 63: ContentWarningNode

uobtheatre/productions/test/test_forms.py
```diff
@@ -1,12 +1,71 @@
...
+
+    form.save()
+    assertQuerysetEqual(production.content_warnings.all(), [])
+
+
+@pytest.mark.django_db
+def test_production_form_null_warnings():
+    production = ProductionFactory()
+    warning = ContentWarningFactory()
+    ProductionContentWarning.objects.create(warning=warning, production=production)
```
JElgar, line 53: We should have a factory for this

schema.graphql
```diff
@@ -552,8 +569,10 @@ type ProductionNode implements Node {
   status: EnumNode
   ageRating: Int
   facebookEvent: String
-  warnings: [WarningNode!]
+  supportEmail: String!
+  contentWarnings: [ProductionContentWarningNode!]
   slug: String!
+  warningsPivot(offset: Int, before: String, after: String, first: Int, last: Int): ProductionContentWarningNodeConnection!
```
JElgar, line 575: Do we want to expose this?

JElgar (commented):
We should not merge things without reviews especially when they include migrations as reverting those is a mega pain.  

Not sure what we want to do with this pr there seem to be some issues with it but nothing major. 
Support email is my biggest gripe so we should catch up on that.

PR #554, opened 2022-07-07
repo: Fix CI

atoff:
Bump stubs version used in pre commit to fix CI

Can un-fix stubs version once https://github.com/typeddjango/django-stubs/pull/1030 is merged (probably)


JElgar (commented):
Is it required that these versions are different?

atoff:
> Is it required that these versions are different?

good point, updated

PR #592, opened 2022-09-03
feat: implement production excerpt

Guidedriver0:
Alex made me do a thing


Guidedriver0:
Re: this mess:

https://github.com/BristolSTA/uobtheatre-web/pull/277

uobtheatre/productions/schema.py
```diff
@@ -223,6 +225,9 @@ def resolve_total_tickets_sold(self, info):
     def resolve_content_warnings(self, info):
         return self.warnings_pivot.order_by("warning__short_description").all()
 
+    def resolve_excerpt(self, info):
```
JElgar, line 228: Graphene should automatically use properties as resolvers so you should be able to remove this function

uobtheatre/productions/migrations/0025_auto_20220903_1859.py
```diff
@@ -0,0 +1,41 @@
...
+        ("productions", "0024_merge_20220808_2050"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="production",
+            name="excerpt_text",
+            field=models.CharField(blank=True, max_length=255, null=True),
+        ),
+        migrations.AlterField(
```
JElgar, line 20: There are a few migrations here that shouldn't be in this pr. This isnt your fault but we should fix this elsewhere so let's not merge this until weve added the missing migrations to dev. Once we've done that you will have to remove this migration and regenerate.

uobtheatre/productions/models.py
```diff
@@ -979,6 +982,14 @@ def total_tickets_sold(self) -> int:
             performance.total_tickets_sold() for performance in self.performances.all()
         )
 
+    @property
+    def excerpt(self) -> Optional[str]:
```
JElgar, line 986: excerpt feels like a slightly weird name for this? Is this not just a short description?

JElgar (commented):
Looks great couple of minor comments.

We will also need to wait for a pr to fix some issues after the work on warnings but we can sort that out tomorrow.

uobtheatre/productions/models.py
atoff: https://www.google.com/search?q=excerpt - This is the common name given to a short-version of a description like this

JElgar: Yeah that definition is not what we are using it for. An excerpt is text extracted directly from the text. It's not a summary or description of it. Unless the idea is to let production teams add an excerpt from the script (in which case having the default value being the description makes no sense) this name is wrong.

Guidedriver0: What do u want to call it then?

JElgar: description_short

Guidedriver0:
I have done another thing
Change name to short decription

uobtheatre/productions/models.py
```diff
@@ -980,6 +983,14 @@ def total_tickets_sold(self) -> int:
             performance.total_tickets_sold() for performance in self.performances.all()
         )
 
+    @property
+    def shortdescription(self) -> Optional[str]:
```
atoff, line 987: Can we do what we did previously with `excerpt_text` and `excerpt`?

The field on the model should be `short_description_text` and this property should be `short_description`. Looks nicer than all one word

uobtheatre/productions/schema.py
```diff
@@ -223,6 +225,9 @@ def resolve_total_tickets_sold(self, info):
     def resolve_content_warnings(self, info):
         return self.warnings_pivot.order_by("warning__short_description").all()
 
+    def resolve_shortdescription(self, info):
+        return self.shortdescription
+
```
atoff, line 230: Can you try without these lines in this file (and lines 196-197)? According to James we shouldn't need (I'm not too sure but if we don't need it can get rid of it!)

atoff (changes requested):
Last change and then I think we will be good to merge! Just nice to be super clear and consistent with naming things!

When you change the names, you will need to delete the migration added in this PR (0026_production_short_description) and regenerate it with `make migrations`

uobtheatre/productions/models.py
atoff: @Guidedriver0 can we just axe this completely

atoff (approved):

PR #601, opened 2022-09-22
Release

atoff:
Features:
- Add concept of Financial Transfers (#548)
- Implement new content warning system (custom pivot model) (#549 and #591)

Maintainance:
- Various package upgrades


JElgar (approved):
We need to move financial transfers out of the payments package. Really we should do this now but neither of us can so... LGTM

PR #606, opened 2022-09-25
feat: Consider discounts in min ticket price

### LINKED ISSUE #418: Min Ticket Price
Fix the min ticket price, so that it returns the min ticket price based on single concessions too

streth11:
closes #418

### DIFF
--- a/uobtheatre/productions/models.py
+++ b/uobtheatre/productions/models.py
@@ -1,6 +1,7 @@
 # pylint: disable=too-many-public-methods,too-many-lines
 import datetime
 import math
+from itertools import product
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
 from autoslug import AutoSlugField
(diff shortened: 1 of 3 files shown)


uobtheatre/productions/test/test_models.py
```diff
@@ -688,6 +688,28 @@ def test_performance_min_price():
     assert performance.min_seat_price() == 10
```
streth11, line 688: is this test still necessary? - i dont think so

uobtheatre/productions/models.py
```diff
@@ -572,9 +573,22 @@ def min_seat_price(self) -> Optional[int]:
...
+            [
+                math.ceil(discount_multiplier * psg_price)
+                for discount_multiplier, psg_price in product(
+                    [1 - discount for discount in self.single_discounts_map.values()]
+                    if len(self.single_discounts_map.values()) > 0
+                    else [1],
+                    [psg.price for psg in self.performance_seat_groups.all()]
+                    if len(self.performance_seat_groups.all()) > 0
+                    else [1],
+                )
```
streth11, line 589: honestly fight me how do i do this nicer

JElgar (approved):
LGMT

uobtheatre/productions/test/test_models.py
```diff
@@ -678,14 +678,35 @@ def test_performance_str(name, start, string):
...
+    DiscountRequirementFactory(discount=null_discount, number=1, concession_type=adult)
+
+    single_discount = DiscountFactory(percentage=0.27, name="child")
+    DiscountRequirementFactory(
+        discount=single_discount, number=1, concession_type=child
+    )
+    null_discount.performances.set([performance])
+    single_discount.performances.set([performance])
+
+    assert performance.min_seat_price() == math.ceil((1 - 0.27) * 10)
```
atoff, line 709: Personally I think these kind of assertions should be explicit and use as little function/complication as possible. Having code and not absolutes in assertions makes it harder to understand unit tests like these, as well as introduces the possibility the the output it not actually what you want (because you just copied the logic from the function in the model).

```suggestion
    assert performance.min_seat_price() == 8
```

atoff (changes requested):
One click to accept my suggestion, but otherwise this looks good!

atoff (approved):

atoff:
@streth11 Do you want to create a quick PR to modify the language shown on the frontend? (i.e. no longer excludes concessions, but does exclude fees) https://github.com/BristolSTA/uobtheatre-web/blob/develop/components/production/ProductionBanner.vue#L88

PR #607, opened 2022-09-26
feat: User and time storage for checking in tickets

### LINKED ISSUE #596: Convert check in tickets boolean to timestamp and user
Replace the current "checked_in" boolean with a "checked_in_at" timestamp and "checked_in_by" user.
Also update mutations to work with this

streth11:
closes #596


uobtheatre/bookings/test/factories.py
```diff
@@ -58,6 +58,9 @@ class TicketFactory(factory.django.DjangoModelFactory):
         "uobtheatre.discounts.test.factories.ConcessionTypeFactory"
     )
 
+    # checked_in_at = factory.Faker("future_datetime", tzinfo=timezone.get_current_timezone())
```
JElgar, line 61:
```suggestion
```

uobtheatre/bookings/test/test_models.py
```diff
@@ -877,37 +877,48 @@ def test_booking_pay_deletes_pending_payments():
...
+    ticket = TicketFactory(checked_in_at=initial_check_time)
 
-    assert ticket.checked_in == initial_state
-    ticket.check_in()
-    assert ticket.checked_in == final_state
+    assert ticket.checked_in == initial_check_state
+
+    ticket.check_in(user=gql_client.user)
+    assert ticket.checked_in == final_check_state
+    assert ticket.checked_in_at is not None or not "2000-01-01T00:00:00+00:00"
```
JElgar, line 900: Is this real? This line seems a bit voodoo to me

JElgar: Can you mock the datetime and check its set correctly?

uobtheatre/bookings/schema.py
```diff
@@ -22,6 +22,11 @@ class Meta:
 
 
 class TicketNode(DjangoObjectType):
+    checked_in = graphene.Boolean()
+
+    def resolve_checked_in(self, _):
```
JElgar, line 27: Do you need this?

uobtheatre/bookings/schema.py
```diff
@@ -22,6 +22,11 @@ class Meta:
 
 
 class TicketNode(DjangoObjectType):
```
JElgar, line 24: I think we need to add so auth to the checked_in_by field. I cant remember if there is a nice way of doing that but might just be easiest to do something like
```python
def resolve_checked_in_by(self):
     if () # remind me to try and write this tomorrow 😂
```

JElgar (changes requested):

schema.graphql
```diff
@@ -882,6 +894,7 @@ type UserNode implements Node {
   societies(offset: Int, before: String, after: String, first: Int, last: Int, id: ID, name: String, slug: String, userHasPermission: String): SocietyNodeConnection!
   bookings(offset: Int, before: String, after: String, first: Int, last: Int, createdAt: DateTime, updatedAt: DateTime, status: String, reference: String, user: ID, creator: ID, performance: ID, adminDiscountPercentage: Float, expiresAt: DateTime, id: ID, statusIn: [String], search: String, checkedIn: Boolean, active: Boolean, expired: Boolean, orderBy: String): BookingNodeConnection!
   createdBookings(offset: Int, before: String, after: String, first: Int, last: Int, createdAt: DateTime, updatedAt: DateTime, status: String, reference: String, user: ID, creator: ID, performance: ID, adminDiscountPercentage: Float, expiresAt: DateTime, id: ID, statusIn: [String], search: String, checkedIn: Boolean, active: Boolean, expired: Boolean, orderBy: String): BookingNodeConnection!
+  ticketsChecked(offset: Int, before: String, after: String, first: Int, last: Int): TicketNodeConnection!
```
streth11, line 897: restrict this just to the ticket object and not deeper?

atoff: I think this could be a bit clearer - is it checked tickets that belong to the user, or tickets that have been checked in by the user. But also I think maybe lets not expose this for now? I don't see us needing this any time soon.

uobtheatre/bookings/models.py
```diff
@@ -719,18 +728,29 @@ def seat_price(self) -> int:
...
 
-    def check_in(self):
+    @property
+    def checked_in(self) -> bool:
+        """Boolean property for if a ticket is checked in.
+
+        Returns:
+            bool: Whether the ticket is checked in.
+        """
+        return self.checked_in_at is not None
```
atoff, line 738: Do we need this? If `checked_in_at` has a timestamp, it will evaluate as truthy, if it doesn't it will evaluate as falsey.

I.e. we can do `if ticket.checked_in_at:` and have the same effect

uobtheatre/bookings/test/factories.py
```diff
@@ -58,6 +58,8 @@ class TicketFactory(factory.django.DjangoModelFactory):
         "uobtheatre.discounts.test.factories.ConcessionTypeFactory"
     )
 
+    checked_in_by = factory.SubFactory(UserFactory)
+
```
atoff, line 62: By default this will mean any ticket we create via a factory is partially checked in (because they have a `checked_in_by` but not a `checked_in_at`. We either need to set both or neither to comply with our logic. I would side on default to not having tickets checked in from a factory, to I would remove this line. (I realise this means you have to provide one when you create in your tests - @JElgar might be able to chime in about if we can have optional states for our factories that would do this for you)

uobtheatre/bookings/test/test_models.py
```diff
@@ -877,37 +877,53 @@ def test_booking_pay_deletes_pending_payments():
...
-    ticket = TicketFactory(checked_in=initial_state)
+        assert ticket.checked_in == initial_check_state
 
-    assert ticket.checked_in == initial_state
-    ticket.check_in()
-    assert ticket.checked_in == final_state
+        ticket.check_in(user=gql_client.user)
+        assert ticket.checked_in == final_check_state
+        assert ticket.checked_in_at == mock_ticket_check_in_time
+        assert ticket.checked_in_by == gql_client.user
```
atoff, line 904: Not quite sure why you are using `gql_client` in a model unit test - it looks like just to get a user?
Just use `user = UserFactory()` instead.

uobtheatre/bookings/test/test_mutations.py
```diff
@@ -2463,7 +2463,7 @@ def test_check_in_booking_fails_if_not_paid(gql_client, status):
         performance=performance, user=gql_client.user, status=status
     )
 
-    checked_in_ticket = TicketFactory(booking=booking, checked_in=True)
+    checked_in_ticket = TicketFactory(booking=booking, checked_in_at=timezone.now())
```
atoff, line 2466: Part of the mutation logic that you introduce with this PR is that you pass the current logged in user as the user who checked in the ticket (https://github.com/BristolSTA/uobtheatre-api/pull/607/files#diff-6d10df12e83f25dd8b0a9a9df90593da84dd0a31699625bb5aec191948fb6bdcR350). This isn't tested anywhere in this file. You should add an assertion in the `test_check_in_booking` (https://github.com/BristolSTA/uobtheatre-api/blob/263b195337eea082f872ce158180a7d206d7ee57/uobtheatre/bookings/test/test_mutations.py#L2284) method that the current GQL user is set as the checked_in_by user on the ticket

atoff (changes requested):
Generally looks fab, just a few small things to fix up!

uobtheatre/bookings/models.py
```diff
@@ -681,7 +683,14 @@ class Ticket(models.Model):
     )
     seat = models.ForeignKey(Seat, on_delete=models.RESTRICT, null=True, blank=True)
 
-    checked_in = models.BooleanField(default=False)
+    checked_in_at = models.DateTimeField(null=True, blank=True, default=None)
+    checked_in_by = models.ForeignKey(
+        User,
+        on_delete=models.RESTRICT,
+        related_name="tickets_checked_in_by_me",
```
atoff, line 690:
```suggestion
        related_name="ticked_checked_in_by_user",
```
Then delete and regenerate your migrations

atoff (changes requested):

uobtheatre/bookings/test/factories.py
```diff
@@ -58,9 +59,18 @@ class TicketFactory(factory.django.DjangoModelFactory):
...
     )
 
+    checked_in_by = None
+    checked_in_at = None
+
     class Meta:
         model = Ticket
 
+    class Params:
+        create_checked_in = factory.Trait(
```
JElgar, line 69:
```suggestion
        set_checked_in
```

uobtheatre/bookings/models.py
```diff
@@ -719,18 +728,29 @@ def seat_price(self) -> int:
...
+    @property
+    def checked_in(self) -> bool:
+        """Boolean property for if a ticket is checked in.
+
+        Returns:
+            bool: Whether the ticket is checked in.
+        """
+        return self.checked_in_at is not None
+
+    def check_in(self, user: User):
```
JElgar, line 740: This should raise an exception in the event of a ticket being checked in. We probably already handle this exception in the resolver but can we move this here?

uobtheatre/bookings/models.py
```diff
@@ -719,18 +728,29 @@ def seat_price(self) -> int:
...
         """
         Check a ticket in
         """
-        self.checked_in = True
+        self.checked_in_at = timezone.now()
+        self.checked_in_by = user
         self.save()
 
     def uncheck_in(self):
```
JElgar, line 748: I guess same here? You can't uncheck in a booking that is not checked in

schema.graphql
```diff
@@ -882,6 +894,7 @@ type UserNode implements Node {
   societies(offset: Int, before: String, after: String, first: Int, last: Int, id: ID, name: String, slug: String, userHasPermission: String): SocietyNodeConnection!
   bookings(offset: Int, before: String, after: String, first: Int, last: Int, createdAt: DateTime, updatedAt: DateTime, status: String, reference: String, user: ID, creator: ID, performance: ID, adminDiscountPercentage: Float, expiresAt: DateTime, id: ID, statusIn: [String], search: String, checkedIn: Boolean, active: Boolean, expired: Boolean, orderBy: String): BookingNodeConnection!
   createdBookings(offset: Int, before: String, after: String, first: Int, last: Int, createdAt: DateTime, updatedAt: DateTime, status: String, reference: String, user: ID, creator: ID, performance: ID, adminDiscountPercentage: Float, expiresAt: DateTime, id: ID, statusIn: [String], search: String, checkedIn: Boolean, active: Boolean, expired: Boolean, orderBy: String): BookingNodeConnection!
+  ticketsCheckedInByUser(offset: Int, before: String, after: String, first: Int, last: Int): TicketNodeConnection!
```
JElgar, line 897: This is slightly concerning. Currently we have a UserNode and an ExtendedUserNode. This field should be private in most cases but we havent set the permissions for it. This is fine for now as UserNode is only returned from ObtainJWT but when we phase out the current auth package (or when we add a user to any response) we need to make sure we keep this in mind

atoff (approved):

uobtheatre/bookings/models.py
```diff
@@ -681,7 +684,14 @@ class Ticket(models.Model):
     )
     seat = models.ForeignKey(Seat, on_delete=models.RESTRICT, null=True, blank=True)
 
-    checked_in = models.BooleanField(default=False)
+    checked_in_at = models.DateTimeField(null=True, blank=True, default=None)
```
JElgar, line 687: deafult=None feel like it should be deafult? Do we need this?

JElgar (approved):
LGTM

PR #644, opened 2023-01-20
schema: Make booking fields required

atoff:


JElgar (approved):
Oh wow, now this is wonderful

PR #662, opened 2023-02-24
Adding discount perm

Guidedriver0:
Adding apply_booking_discount


atoff:
@Guidedriver0 a few tests to fix :)

Guidedriver0:
Yeah will do now

.docker/canary.json
```diff
@@ -0,0 +1 @@
+Get "https://desktop.docker.com/compose-canary.json": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```
atoff, line 1: Delete this file

uobtheatre/productions/migrations/0027_alter_production_options.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('productions', '0026_auto_20221210_2056'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
```
atoff, line 13: So I was thinking it would be a good idea to give existing production teams this new permission. We could either do this manually once deployed or we can do it the proper way, with a mutation in this migration so that after it creates the permission, it gets all the users with explicitly assigned "change_production" permissions and grants them the new permission. What do you think?

An example of where we did this previously; when we added the check in timestamp to bookings: https://github.com/BristolSTA/uobtheatre-api/blob/dev/uobtheatre/bookings/migrations/0006_alter_ticket_checkin_to_datetime.py

atoff (changes requested):

uobtheatre/productions/migrations/0027_alter_production_options.py
Guidedriver0: But we don't want everybody with "change_production" permission to have the new permission (like SM's and the approval team). would it not be easier to do it manually or could we find a way to make a distinction?

atoff: We'd only apply it to people who have "change_production" applied explicitly to the object (this equates to the production team at the moment, because other people have the global permission through groups instead). Essentially, we would be grandfathering in the fact that people who currently have change_production applied to a specific production should be able to create a comp booking. I don't mind doing it manually, but need to make sure we remember to do this when it goes live

uobtheatre/productions/migrations/0027_alter_production_options.py
```diff
@@ -0,0 +1,29 @@
...
+from django.db import migrations, models
+
+
+def add_apply_booking_discount(apps, schema_editor):
+    Permission = apps.get_model('auth', 'Permission')
+    Group = apps.get_model('auth', 'Group')
+    apply_booking_discount = Permission.objects.get(codename='apply_booking_discount')
+    change_production = Permission.objects.get(codename='change_production')
+    for group in Group.objects.filter(permissions=change_production):
+        group.permissions.add(apply_booking_discount)
```
atoff, line 14: Should also find users who have the permission, not just groups? Apart from that I think this seems to do what it should do, but we should do a quick run locally to make sure

uobtheatre/productions/migrations/0027_alter_production_options.py
```diff
@@ -0,0 +1,29 @@
...
+
+    dependencies = [
+        ('productions', '0026_auto_20210224_1911'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="production",
+            name="apply_booking_discount",
+            field=models.BooleanField(blank=False, null=False),
```
atoff, line 26: This seems to be making a `apply_booking_discount` field on the `Production` model for some reason. Is this correct?

PR #695, opened 2023-03-28
feat: Better Sentry Logging

atoff:


Makefile
```diff
@@ -154,12 +148,11 @@ mypy: ## Type checking - mypy
 schema: ## Dumps graphql schema in schema.json
 	$(COMMAND_PREFIX) ./manage.py graphql_schema --schema uobtheatre.schema.schema --out schema.graphql
 
-pr: ## Runs everything required (that is not included in precommit) for a pr
+pr: ## Runs everything required for a pr
```
JElgar, line 151: Agree precommit is annoying and we should remove it but we should add back the functionality it provided here. It did all sorts.but most importantly lint and mypy

uobtheatre/addresses/migrations/0001_initial.py
```diff
@@ -4,7 +4,6 @@
 
 
 class Migration(migrations.Migration):
-
```
JElgar, line 7: What linting rules are being applied now?

atoff: The same pylint ones. I don't think these files have ever been touched because you force committed...

Makefile
atoff: The functionality is already included in make lint?

JElgar (approved):

PR #716, opened 2024-01-29
Fix the make seed command

zst-c:
Previously, `make seed` wasn't working, because of updates to the database schema that was not reflected in the various `fixture.json` files. This is now remidied.


Guidedriver0 (approved):
nice

PR #719, opened 2024-04-12
Add a production filter to the bookings() query

zst-c:
### What?

- This adds three filters to a `bookings` query.
  1. Based on the `name` of a production linked to each booking.
      - This filter is **whitespace sensitive**, so that a query for e.g. `"A Show"` would only match `"A Show"` and **not** `"A" || "Show"`.

  2. Based on the `slug` of a production linked to each booking. This must match exactly.

  4. Based on the `performanceId` of a performance linked to each booking. This must also match exactly.
      - This ID is in base64 format, and converted to the corresponding numeric ID within the query.

- This also adds tests for the various new filters.

### Why?

- This is used by the global bookings page in the admin part of the Web repo. (See [this PR](https://github.com/BristolSTA/uobtheatre-web/pull/375)).

### Notes

- We should consider if we want the first filter just to search the production `name`, or if we want it to search the `slug` or anything else as well.

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -274,8 +274,8 @@ type ExtendedUserNode {
   dateJoined: DateTime!
   id: String
   email: String!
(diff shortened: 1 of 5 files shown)


HappyNTH (approved):
All looks good to me, thank you very much!

PR #814, opened 2024-09-19
feat: Site-wide maintenance/warning messages

ClaraARay:
## Description

Add a new table to the database for Site Messages, intended to be displayed as global warning/information/maintenance alerts on the front-end where required (e.g. to give advanced warning of site downtime, or to provide urgent updates). Site messages include a event start time, an event end time (or the option to set them as indefinite), a display start time, a boolean to mark them active, a type (for a set list - `Maintenance`, `Alert`, and `Information`), a free-text message, and are tied to a creating user. Django is also set up to allow the creation/editing them through the admin panel, since the first web implementation will not include a front-end system to create and edit them.

The whole site message node can be queried (with a variety of filters) or a specific entry by its ID. There is also a mutation to create or update them, once web implementation accompanies it, which is restricted behind permissions for creation, editing & deleting. 

A suite of tests have also been written to cover all additions.

Will accompany https://github.com/BristolSTA/uobtheatre-web/pull/565

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ check-users: ## Do the migrations
 	$(COMMAND_PREFIX) `python manage.py number_of_users | tail -n 1` | grep 0
 
 seed: ## Seed the db with some example data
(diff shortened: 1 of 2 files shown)


uobtheatre/site_messages/__init__.py
zst-c, line 1: I'm assuming this file does need to exist, even though it's literally entirely empty?

zst-c (approved):
Lots of good tests, thank you!

PR #820, opened 2024-09-30
feat: Implement Turnstile

HappyNTH:


HappyNTH:
Turnstile presently stands as its own mixin, and could easily be mixed into other mutations that send emails (password reset, change email, resent activation email) and that should be protected against brute force (logins). Logins need extra care though, as users need to use the `login` mutation to obtian a request header is working with graphqli.

config/settings/common.py
```diff
@@ -114,6 +114,11 @@
     }
 DATABASES["default"]["ATOMIC_REQUESTS"] = True
 
+# Turnstile
+TURNSTILE_SECRET = env(
+    "TURNSTILE_SECRET", default="1x0000000000000000000000000000000AA"
```
zst-c, line 119: Fine to merge as-is; I wonder in the future though, whether we should error with a sensible message if there's no secret in the `.env` file, rather than just falling silently to a nonsense default.

schema.graphql
```diff
@@ -345,7 +345,7 @@ type Mutation {
   deleteDiscount(id: IdInputField!): DeleteDiscountMutation
   discountRequirement(input: DiscountRequirementMutationInput!): DiscountRequirementMutationPayload
   deleteDiscountRequirement(id: IdInputField!): DeleteDiscountRequirementMutation
-  register(email: String!, firstName: String!, lastName: String!, password1: String!, password2: String!): Register
+  register(turnstileToken: String, email: String!, firstName: String!, lastName: String!, password1: String!, password2: String!): RegisterTurnstile
```
zst-c, line 348: Technically, this should be `turnstileToken: String!`, as it's a required (read: non-null) argument.

zst-c (approved):
All looks good here - left a couple of tiny pedantic comments, but they very much don't need fixing; it's only if we want to, but more than happy to merge as-is.

schema.graphql
zst-c: Ah yes - looks like [here](https://github.com/BristolSTA/uobtheatre-web/pull/641/files#diff-204ad7265ed9910b0b2654a7d9faccfe7bd432c8d4b4dc35c1534b972cb1f7cdR7) in your code changes, we are defining it as non-null.

config/settings/common.py
HappyNTH: The default fall-back is one of Cloudflare's testing keys, which always produces a pass. It means that if people are developing elsewhere they don't have to worry about putting in testing keys, and can just happily fall back to the default.

zst-c: So if we accidentally forgot to set the environment variable, or for some reason it can't find it, we'll automatically basically silently accept all captcha responses?

PR #833, opened 2024-10-10
fix: Allow those with permission to approve productions to reject them too

### LINKED ISSUE #822: STA users with approval permissions cannot reject productions

HappyNTH:
A very simple fix to resolve #822, and to add a test to cover the scenario.


zst-c (approved):
Perfect, no notes

PR #846, opened 2024-10-21
feat: relaxed performance update

finn768:
API additions for relaxed performance info on frontend.


HappyNTH:
First of all, thank you so much for the contribution! I'm glad you got everything working, and I look forward to working with you in the future on more features.

Now, specific things as we discussed yesterday. I like having a way to demarkate a relaxed performance, but the definition of a relaxed performance varies between societies and productions, so we need a way to reflect that. Some societies, for example, include a character introduction in their relaxed performance, while others just have house lights up. To make sure we're giving customers the right information, we should reflect this granularity.

One way to do this might be to use a system similar to content warnings, whereby we offer production companies a few set phrasing options (e.g. `House lights will remain partially up, loud noises will be reduced and some special effects will be removed` vs, `Characters will be introduced, ....`) and also give an option to enter a custom string instead. That being said, I can see a couple of issues here:
1. Lots of the strings will probably be very similar to one another. Is it therefore perhaps worth having a set of tick boxes for a production company to include (`House lights will remain partially up`, loud noises will be reduced`, `cast will introduce their characters`, `pyrotechnic effects will not feature`, etc.) that they can pick and choose from, alongside a custom field for further information?
2. Is this potentially just a duplication of the information that is put in the `Performance Description` field? As it currently stands it is there where production companies add information about relaxed performances, and so potentially just showing that field earlier in the booking flow (like https://github.com/BristolSTA/uobtheatre-web/issues/284) will expose all the information we need. _However_, if we give a bit more detail to relaxed performances - and can come up with a good way to show both of these on the frontend - I see value in seperating it from the performance description

Sorry there's a lot of thinking here - drop me a message if you want to talk this through further, I'm sure we can meet up and chat! This is definitely a good idea overall, and thanks for taking a crack at it, we just need a more nuanced approach.

ClaraARay:
See #967

PR #918, opened 2025-02-27
feat: Add a booking accessibility information mutation

HappyNTH:
Adds a mutation to allow a booking's accessibility information to be edited aside from the main Booking mutation. The specific mutation ensures that an email is sent when things are updated.


HappyNTH:
Should we procedurally delete accessibility information when a production is completed?

uobtheatre/addresses/models.py
```diff
@@ -13,6 +13,12 @@ class Address(models.Model):
     street = models.CharField(max_length=255)
     city = models.CharField(max_length=255)
     postcode = models.CharField(max_length=9)
+    what3words = models.CharField(
+        max_length=255,
```
zst-c, line 17: For info, I believe the (max W3W address length is 64 chars)[https://developer.what3words.com/design/maker-guidelines#the-basics]

uobtheatre/bookings/emails.py
```diff
@@ -85,3 +85,71 @@ def send_booking_accessibility_info_email(
...
+        .values_list("email", flat=True)
+    )
+
+    mail = (
+        MailComposer()
+        .greeting()
+        .line(
+            f"A booking for {booking.performance} has had its accessibility information updated. The new accessibility information is:"
+        )
+        .line(f"'{booking.accessibility_info}'")
```
zst-c, line 144: Similar questions about sending data.

uobtheatre/productions/mutations.py
```diff
@@ -144,6 +144,17 @@ def resolve_mutation(
...
 
+        elif (
+            status == Production.Status.COMPLETE
+            and previous_status == Production.Status.CLOSED
+        ):
+            # Clear any involved bookings of accessibility information
+            for performance in production.performances.all():
+                for booking in performance.bookings.all():
+                    if booking.accessibility_info:
+                        booking.accessibility_info = None
```
zst-c, line 155: Do you need to also update the `booking.has_accessibility_info` field?

zst-c (commented):
Looks good to me. I've added some thoughts, rather than necessary changes - happy to approve, just want to make sure they've been considered.

uobtheatre/bookings/models.py
```diff
@@ -201,6 +201,21 @@ def expired(self, bool_val=True) -> QuerySet:
...
+
+        Args:
+            bool_val (bool): when True: return only bookings with accessibility information,
+            when False: return only bookings without accessibility information
+
+        Returns:
+            QuerySet: the filtered queryset
+        """
+        if bool_val:
+            return self.exclude(accessibility_info__isnull=True)
```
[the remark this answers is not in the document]
HappyNTH: It can be simplified to `return self.exclude(accessibility_info__isnull=bool_val)`, as [the speed difference seems negligable](https://stackoverflow.com/questions/2422380/django-filter-vs-exclude). It needs to rest on `bool_var` though so `has_accessibility_info(false)` can be checked in the query.

uobtheatre/bookings/mutations.py
```diff
@@ -112,6 +114,57 @@ class Meta:
...
+            )
+        return super().authorize_request(_, info, **inputs)
+
+    @classmethod
+    def resolve_mutation(cls, _, info, booking_id, accessibility_info):
+        booking = Booking.objects.get(id=booking_id)
+        previous_accessibility_info = booking.accessibility_info
+
+        # Check the booking is in the future, otherwise return error
+        if booking.performance.start and booking.performance.start < timezone.now():
```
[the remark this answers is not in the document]
HappyNTH: Timezones are just the worst to try and code around when daylight savings is mixed in. 

But I don't think this will bite us, since `timezone.now()` is evaluating against django's timezone, not the client's, as `performance.start` is set using.

PR #927, opened 2025-03-19
feat: Add a comp ticket permission

HappyNTH:
Adds a new production permission (both global and production-specific) to create complimentary tickets, grantable by anyone who has permission to change the production.

Users with those permissions can now create complimentary tickets when a production is approved, but not yet live. This is to fix an issue Dance & MTB 48hr have run into where they want to comp tickets before making a production live to avoid selling out. 

In doing so, this PR closes a slight loophole where users with just box office permissions could theoretically create comp tickets, and the bug on the other side where users with no box office permissions but still change production permissions could not create comp tickets because they could not create tickets for other users.

Also tightens up some tests to check for when a user can book.


zst-c (approved):
This looks great, thank you!

PR #955, opened 2025-05-14
fix: Async Refunds & Square Webhooks

ClaraARay:
I made an oversight on my partial refunds PR and missed including fees-accomodating refunds in async refunds (our primary type), so this is a fix for that!

Also, when Square informs us a refund has been completed (via webhook), the booking is now marked as `Refunded` at the same time the transaction is completed.


zst-c (approved):
A small PR, and all looks good, so happy to approve with no changes!

PR #967, opened 2025-07-06
feat: relaxed performance option

ClaraARay:
Mirrors #846, but brings it onto the repository so it can have access to necessary keys

API additions for relaxed performance info on frontend


uobtheatre/productions/test/factories.py
```diff
@@ -46,6 +47,8 @@ class PerformanceFactory(factory.django.DjangoModelFactory):
     disabled = False
     production = factory.SubFactory(ProductionFactory)
     venue = factory.SubFactory(VenueFactory)
+    is_relaxed = True
```
zst-c: Should all test performances be relaxed?? Or should this randomly choose between True and False?

zst-c (changes requested):
I'm happy with this in the round - but don't merge it in until the associated Web PR is approved, in case that requires any changes on the backend.

Alongside the comments left, if you could update the seed data in `fixtures.json` (currently in `dev`, you will need to merge that back in to this branch first!) to reflect the newly added fields, that would be excellent - so that we can keep the fixtures useful as they grow in complexity over time.

Just a note about testing, and obviously you need to make sure you're passing all the tests (including linting!). Just running `make lint` locally, and then fixing any errors that appear should be all you need to do to ensure it passes. (You can also run `make pr` for the full workup).

### TODO:

- [x] help text on the backend
- [ ] `fixtures.json`
- [x] `is_relaxed` field in tests
- [x] `make lint`

uobtheatre/productions/models.py
```diff
@@ -159,6 +159,16 @@ class Meta:
         ordering = ["id"]
 
 
+class RelaxedCategory(models.Model):
+    """Categories of adaptations for relaxed/sensory friendly etc. performances."""
+
+    short_description = models.CharField(max_length=255)
```
zst-c: Please can you add some helper text on the backend so that it shows an explanation of these categories, and crucially of what the `default_` fields do, as well as where the `long` and `short` descriptions will be shown respectively?

Image

You can use the `site_message` objects as an example:

Image

zst-c:
@finn768 is this ready for re-review or should I hold off?

finn768:
Yep, I think this is ready :)

uobtheatre/productions/fixtures.json
zst-c, line 1: I think when you updated the `fixtures.json` files, specifically this one didn't quite work out correctly - it's kept in it old data like `Example show 2`, which shouldn't really be there ideally! In fact, you can actually see how severely this proposes to differ from `dev` based on how big the diff is (3K lines!).

Image

Other than the relaxed performance info, this file should be unchanged from `dev`, and so the diff of this in the PR should be pretty small in the end, adding only a few things.

What would be best to do would be to delete the file locally completely, use `git checkout` to checkout just this singular file from `dev` (`git checkout dev -- uobtheatre/productions/fixtures.json`), and then to tweak it to add in the relaxed performance stuff.

I know it's a bit annoying to do, but in the long term this'll keep our test data clean, and going forwards this will be much easier to do, as all of our branches will be branched off `dev` which will have the correct `fixtures.json` files in (it's just that this branch was obviously made before the changes).

zst-c (changes requested):
So so so almost there, and thank you for making those changes, that's excellent! Especially a fan of the help text and the modifications to the testing, thank you! 😄 

The only thing is about the `fixtures.json` file - that hopefully should be relatively quick to fix, and you'll know when it's working because the diff will be nice and small (and of course, you can test it by spinning up a local version of the api and web containers to test). Once that's done, both happy and excited to merge!

ClaraARay:
@finn768 please do @ me once you've made the final changes, and then I'll merge this (and the web PR) so we can roll out 1.3!

finn768:
@ClaraARay all should be sorted now :)

PR #984, opened 2025-09-30
feat: Further Accessibility Tweaks

ClaraARay:
Required companion to [uobtheatre-web PR 353](https://github.com/BristolSTA/uobtheatre-web/pull/353). Tweaks the venue accessibility field to use TipTapText for rich text presence, adds a `accessibility_short` to the venue model to provide a short accessibility summary to provide limited information on the frontend, and add `can_modify_accessibility` to the booking node to allow web to check if a user can edit a booking's accessibility (rather than having to mirror the checks on the frontend, potentially getting things out of sync)


uobtheatre/productions/models.py
```diff
@@ -812,13 +812,14 @@ def user_can_see(self, user: "User"):
...
         # Productions the user has tickets for that are within the last week or the future
-        productions_user_has_tickets = []
         if user.is_authenticated:
             one_week_ago = timezone.now() - datetime.timedelta(days=7)
             productions_user_has_tickets = self.filter(
                 performances__bookings__user=user,
                 performances__start__gte=one_week_ago,
             ).values_list("id", flat=True)
+        else:
+            productions_user_has_tickets = []
```
ClaraARay: This is to deak with a mypy flagged issue!

schema.graphql
```diff
@@ -56,6 +56,7 @@ type BookingNode implements Node {
   transactions(offset: Int, before: String, after: String, first: Int, last: Int, type: String, provider: String, createdAt: DateTime, id: ID): TransactionNodeConnection
   expired: Boolean!
   salesBreakdown: SalesBreakdownNode
+  canModifyAccessibility: Boolean
```
zst-c: Is there any reason that this isn't mandatory? As far as I can see, this is set [here](https://github.com/BristolSTA/uobtheatre-api/blob/9f1f7e6ece325792640d953751ed828709f5cb38/uobtheatre/bookings/schema.py#L357) and so will always be T or F. But aware I may be missing something. If not, I think we should make it mandatory, with the general concept of "stricter checking is always better".

zst-c (approved):
Happy with all of this to be merged - the comment above is a very tiny thing, so I've pre-emptively approved this if we decide it's not worth the single line change.

schema.graphql
ClaraARay: Nope, no reason for this not to be required!

ClaraARay:
@zst-c I'll merge this in after you've merged in your seed fixes, so I can add accessibility shorts to some of the fixtures

PR #986, opened 2025-10-05
Update fixtures.json to add a load of new seed data

zst-c:
As a result of me testing a load of different PRs, this came up as a nice-to-have, hence it is implemented:

This updates all the various `fixtures.json` files to add a huge amount of new seed data. In the long-term, this will make testing significantly easier, as the website will have proper seeded data when spun up locally. This should be a big QoL improvement.

When new fields are added in the future, we should probably update these fixtures in the same way, to ensure they remain up-to-date (probably with some help of some ~AI~ friends).

## Testing

Testing this with a normal code review is probably not possible due to the size - the best thing to do is likely to check only the non-`.json`/non-`.jpg` files, and then just to spin up a local instance and see how it looks/if it works as expected when `make seed` is run.


ClaraARay:
Holy guacamole 😮

ClaraARay:
> asks for my review
> I go on to review
> 3 new commits
> pull request changed to draft

my feeling when: 😮

zst-c:
> > asks for my review
> > I go on to review
> > 3 new commits
> > pull request changed to draft
> 
> my feeling when: 😮

As soon as it passes all the tests (I'm hoping in the next few minutes), it'll _actually_ be ready to review!!! I just caught a list-minute test error hahaha

zst-c:
All yours @ClaraARay - I shall do no more touchy touchy

ClaraARay (approved):
Looks good, builds perfectly, thank you so much! Very happy to merge.

While we're here, a quick question: do you think there's some way we could test for fields that don't get included in the db seed file as part of our actions? That way if people make edits to the schema they're forced to update the data to cover those cases. I _imagine_ this probably isn't the case, or at least isn't the case in a way that won't take 20 minutes to run and even more to create, but curious if you'd have any ideas. If there is it's definitely beyond the scope of this PR!

PR #989, opened 2025-10-06
feat/fix: Site Message toDisplay Filter & Updated Helper Text

ClaraARay:
Adds a `toDisplay` filter to the SiteMessageNode, so the frontend can offload filtering to the API (and therefore properly account for indefinite overrides).

Also updates a couple of pieces of helper text to make clear that display times are not displayed on modals.


zst-c (approved):
Super duper, thank you! All very clear and working well (:

PR #998, opened 2025-10-20
Bump Django to version 5.2.7 (LTS)

zst-c:
This comes alongside the web update - [#1044](https://github.com/BristolSTA/uobtheatre-web/pull/1044)

## Description

This bumps Django to the latest non-development version, v5.2.7. In doing so, the Python version must also be bumped to 3.12, alongside bumping the Postgres version to PostgreSQL 14.19.

This works to update certain fields, resolvers, enums, types, and other such things. It's a lot of small tweaks throughout!

This also replaces the current `django-graphql-auth` package with a new and improved one, [currently hosted on my own GitHub repo](https://github.com/zst-c/django-graphene-auth). Over time, this should be merged into the API branch and maintained in there, or the repo should be moved to the STA organisation, and adequate branch protection rules should be applied. Either way, this should be separate to this PR, so as not to muddy the PR with a large number of changed files which are unrelated to the actual spec bump.

## Useful Links

- Python version list: https://devguide.python.org/versions/
===
- Postgres releases list: https://www.postgresql.org/docs/release/
===
- Django releases list: https://docs.djangoproject.com/en/5.2/releases/
- Django deprecation policy: https://docs.djangoproject.com/en/4.1/internals/release-process/#deprecation-policy
- Django deprecation timeline for all releases: https://docs.djangoproject.com/en/5.2/internals/deprecation/
- Django upgrade guide: https://docs.djangoproject.com/en/5.2/howto/upgrade-version/
- 4.0 release notes: https://docs.djangoproject.com/en/5.2/releases/4.0/
- 4.1 release notes: https://docs.djangoproject.com/en/5.2/releases/4.1/
- 4.2 release notes: https://docs.djangoproject.com/en/5.2/releases/4.2/
- 5.0 release notes: https://docs.djangoproject.com/en/5.2/releases/5.0/
- 5.1 release notes: https://docs.djangoproject.com/en/5.2/releases/5.1/
- 5.2 release notes: https://docs.djangoproject.com/en/5.2/releases/5.2/


ClaraARay:
Thank you so much for this: I'll review this shortly, but not expecting major issues.

I think we avoid merging this until 1.3 is out the door, so we have a bit of dev time to test it further and catch any issues that might arise before it goes live.

uobtheatre/productions/emails.py
```diff
@@ -1,12 +1,12 @@
...
+from typing import Any, List, Optional
 
 from uobtheatre.mail.composer import MailComposer
 from uobtheatre.productions.models import Performance, Production
 from uobtheatre.users.models import User
 from uobtheatre.utils.lang import pluralize
 
 
-def send_production_approved_email(user: User, production: Production):
+def send_production_approved_email(user: Any, production: Production):
```
ClaraARay, line 9: Why are we loosening the typing requirement here?

ClaraARay (commented):
Looks good to me, thank you again for all this work. One question about loosening typing, but it's not enough to block a merge.

That being said, let's not merge until 1.3 is out (I've not given explicit appoval as a result)

uobtheatre/productions/emails.py
zst-c: Because the arguments that get passed into it are not technically of the `User` type, I believe they're items from a list of type `List[Any] | Str` (or something weird like that!). But it's being returned from a Django function so I can't fix the typing of that, and Mypy really hates it otherwise...

There are a few instances like this where mypy just doesn't believe our typing at all lol

We could revert this to `User` though, and use `type: ignore` on the lines that call this function and throw errors - happy either way.

ClaraARay:
Merged manually after resolving schema clash by commit 8a93787

PR #1006, opened 2025-10-30
feat: Add User-Friendly Payment Error Messages

ClaraARay:
Adds a short handler to humanise Square API Error Messages for end users, as error details are [not intended to be customer facing](https://developer.squareup.com/docs/build-basics/general-considerations/handling-errors#error-object-properties). This covers the [error codes](https://developer.squareup.com/reference/square/payments-api/create-payment) we are likely to face when attempting to create payments, and otherwise falls back to using the error code. 

Adds tests to keep that sweet, sweet 100% coverage!

### DIFF
--- a/uobtheatre/utils/exceptions.py
+++ b/uobtheatre/utils/exceptions.py
@@ -246,13 +246,14 @@ class SquareException(GQLException):
             "TRANSACTION_LIMIT": "The card issuer has determined the payment amount is either too high or too low.",
             "BAD_EXPIRATION": "The card expiration date is either missing or incorrectly formatted.",
             "CARD_DECLINED_VERIFICATION_REQUIRED": "The payment card was declined with a request for additional verification.",
-            "CHIP_INSERTION_REQUIRED": "The card issuer requires the card to be inserted into a chip reader."
+            "CHIP_INSERTION_REQUIRED": "The card issuer requires the card to be inserted into a chip reader.",
         }
 
-        def get_user_readable_error_message(error: Error) -> str:
-            if error.code in passthrough_error_categories:
-                return user_readable_error_details.get(error.code, "There was an issue processing your payment (%s)" % error.detail)
-            return error.detail
+        def get_user_readable_error_message(error_code: Optional[str]) -> str:
+            return user_readable_error_details.get(
+                str(error_code),
+                "There was an issue processing your payment (%s)" % error_code,
+            )
 
         error = (
             api_error.errors[0]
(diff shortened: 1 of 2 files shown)


uobtheatre/utils/exceptions.py
```diff
@@ -181,12 +182,37 @@ def __init__(self, api_error: ApiError):
         passthrough_error_categories = [
             "PAYMENT_METHOD_ERROR",
         ]
+        # Turn common Square errors into user-friendly messages
+        user_readable_error_details = {
+            "ADDRESS_VERIFICATION_FAILURE": "The card issuer declined the request because the postal code is invalid.",
+            "CARD_EXPIRED": "The card issuer declined the request because the card is expired.",
+            "CVV_FAILURE": "The card issuer declined the request because the CVV value is invalid.",
```
zst-c: So of course it's up to us - but granularity of error comes with a decrease in security. For example, if we tell someone that specifically the CVV of a card is wrong, then it's easier for them to iterate through and try different CVVs until they get it right. Whereas if we just tell them that the "card details are invalid" more generally, and they don't know if it's the card number that's wrong, the CVV,  the postcode, or the exp date, then it's significantly harder for them to just iterate through possibilities until they come up with a valid combination. Does that make sense?

My suggestion would be that we maintain lots of these error messages, but maybe merge together `ADDRESS_VERIFICATION_FAILURE`, `CARD_EXPIRED`, `CVV_FAILURE`, `EXPIRATION_FAILURE`, `INVALID_EXPIRATION`, `INVALID_CARD`, `INVALID_PIN`, `PAN_FAILURE`, `BAD_EXPIRATION` into one more generic error about card details being wrong. Of course though, I'm happy to be disagreed with!

(The same argument could be made around telling people specifically about chip verification being required or the card being out of funds, but that's a less important case imo)

zst-c (commented):
I'm happy with the code itself (nice simple changes!), but I've left a comment around security implications. If we don't think it's an issue, then I'm happy to approve and merge.

uobtheatre/utils/exceptions.py
trebor97351: The bank would catch any fraud like this. 
If the payment processor is giving this granularity in the errors, then they are happy with us knowing this info, and so passing it on to the user.

But it is something to bear in mind with other parts of the system, where we are the ones determining which error types exist.

ClaraARay: I think you're right, Robbie, in that the bank should catch this. Still, I don't think it's necessarily terrible to slim it down, mainly because it's a bit of a needless level of detail.

PR #1011, opened 2025-12-07
fix: Only send emails when there is actually accessibility info.

trebor97351:
Often, users put 'No' or similar when they don't have accessibility requirements. This check will prevent emails being sent for accessibility infos with four or less characters.


ClaraARay (approved):
LGTM

PR #1019, opened 2026-03-19
Integrate Graphql Auth

ClaraARay:
Building on #998, which bumped our django version by way of forking the `django-graphql-auth` package we rely on, this PR brings said package directly into the project. In the process it also ties in relevant unit tests from the package to our system, to (hopefully) make sure nothing slips as we go through django versions.

In the process this also resolves the error within the forked `django-graphql-auth` that prevents our deploy workflow from working.


zst-c (approved):
This all seems to work to me - the tests are running well locally, testing in-browser shows no errors, and the migration seems to be effective!

PR #1020, opened 2026-04-29
feat: allow bristolsta.com to get production information

finn768:
Adds (staging.)bristolsta.com to the domains allowed to make cross-origin requests for production information needed for the "What's On" page.


ClaraARay (approved):
LGTM!

